PR #755, opened 2021-11-01
fix: Flyway script to fix roles

Marcellofabbri:
Update table tcs.Person so that rows have the column 'role' correctly spelt (taken from the reference.Role table, wherever casing is the only difference). It won't fix other errors such as extra dots or typos).

From what I've seen it appears that MySQL automatically does not perform the update if it sees that the value that is being set is the same, which would save us from writing a more elaborate script that checks if the update is necessary or not.

If we discover that our script should indeed have the responsibility to check for this condition itself, an option could be something like:
```
UPDATE tcs.Person person
INNER JOIN reference.Role role ON person.role = role.code
SET person.role = if(binary person.role = binary role.code, role.code, role.code);
```

TIS21-1629

### DIFF
--- a/tcs-service/src/main/resources/db/migration/data/hee/V4.10__fix_person_roles_with_inaccurate_casing.sql
+++ b/tcs-service/src/main/resources/db/migration/data/hee/V4.10__fix_person_roles_with_inaccurate_casing.sql
@@ -1,0 +1,3 @@
+update Person person
+inner join reference.Role role ON person.role = role.code
+set person.role = role.code;


ReubenRobertsHEE (commented):
This looks fine. Just checking that the intention is to update two roles (i.e. not just the one mentioned here: https://hee-tis.atlassian.net/browse/TIS21-1633)? 

The roles that will be updated are 
'Dentist in training' / 'Dentist In Training' -> 'Dentist in Training', and 
'Dr in training' / 'Dr in Training' -> 'DR in Training'

Marcellofabbri:
> This looks fine. Just checking that the intention is to update two roles (i.e. not just the one mentioned here: https://hee-tis.atlassian.net/browse/TIS21-1633)?
> 
> The roles that will be updated are 'Dentist in training' / 'Dentist In Training' -> 'Dentist in Training', and 'Dr in training' / 'Dr in Training' -> 'DR in Training'

Good consideration. Ideally yes, it should act on all roles in a row, but I don't think it will. When matching each person.role to role.code, it will find that the string "DR in Training,Dentist in Training" doesn't match any role, so it will fail to include them in the join table. 
Moreover, I just found there are records whose role is "Dr in Training, DR in Training", so even if we did find a way to fix all roles in the multivalue role we'd then have to clean up duplicates that end up looking like "DR in Training,DR in Training".

Marcellofabbri:
This new script I added is way more elaborate (read: intricate) than the other, but I believe it works. I tried it locally and it seemed to do the job. Even roles that are wedged in between several other roles in a comma separated string get fixed.

**I know** it looks convoluted but I don't think there's any way around it: the Roles need to be deconstructed, fixed, reconstituted, then written back into Person.Role.

The concept behind it is:
- unwind the Person table into a temporary one where each row has one role
```
1 | Dr in training,Dentist in training,yh education | CURRENT | 2020-09-08
2 | Dr in training | INACTIVE | 2016-05-30
becomes
1 | Dr in training | 
1 | Dentist in training | 
1 | yh education |
2 | Dr in training |
```
- update with a join table each role with the value matched from the REFERENCE.Role table
```
1 | Dr in Training | 
1 | Dentist in Training | 
1 | YH Education |
2 | DR in Training |
```
- reconcatenate the roles back up
```
1 | Dr in Training,Dentist in Training,YH Education | 
2 | DR in Training |
```
- update the Person table using the ids to map the new concatenated roles to the rows

Judge40 (changes requested):
While this works locally it won't work in a live environment. The database user used in each service only has access to its own schema, so we're unable to query for Role values from the reference schema when running this script in TCS.
I'd also be concerned about the amount of time it might take to run this, TCS would be down for the duration. We could run it out of hours but any heavy migration like this makes me a touch nervous.

I do think its a clever solution but it's a huge amount of scope creep, the ticket only asks for single role variants of "DR in Training" to be fixed - ignoring multi-value roles (and other single roles) altogether.

ReubenRobertsHEE:
Yes, I see there's a good number of people with multiple roles, some of which are duplicated. It might be worth checking back with the ticket owner to see if fixing this is 'in scope' - to me it seems like it might be a separate bit of work! There might be some checking to see how this came about, and potentially fixing it at source to ensure it doesn't happen again.

ReubenRobertsHEE:
Ah, my comment was made before I saw the work you'd done. Should have refreshed the page!

Marcellofabbri:
It may be out of scope in respect to the ticket itself, but also the ticket may fail to grasp the width of the scope. It feels like a data quality fail to go in, fix some roles, and leave other wrong ones there. We didn't restrict it to 'DR in Training' only in the other two PRs that aim at disallowing wrong codes to sneak in via bulk-upload, so that's where I struggle to see the blurred lines of: "the scope".

If we really only want to act on 'Dr in Training' alone, then I guess this would do:
```
UPDATE tcs.Person person
SET person.role = 'DR in Training'
WHERE UPPER(person.role) = 'DR IN TRAINING';
```

But I'm still convinced we're letting an opportunity to act on other wrong codes go. I'm fine with not addressing it in this ticket specifically, however I think we at least ought to check with the POs/Data Leads to see if they'd see fit the creation of another ticket aimed at tackling the other roles.

> While this works locally it won't work in a live environment. The database user used in each service only has access to its own schema, so we're unable to query for Role values from the reference schema when running this script in TCS.

The issue of not seeing Reference from Tcs could be solved by chucking the roles in a temporary table.

Judge40 (approved):

Judge40 (approved):

PR #756, opened 2021-11-02
feat: Change PersonValidator

Marcellofabbri:
**TIS21-1629: Update bulk person create to validate and set the role correctly**

Make checkRole() in PersonValidator use the new endpoint from REFERENCE that returns a Map<String,String> instead of a
Map<String,Boolean>.

E.G.:
- role `dr in training` is passed in
- the `/roles/matches/` Reference endpoint will return a Map like: { "dr in training" : "DR in Training" }.
- the value (retrieved from the DB, therefore correctly cased) is used to set `personDto.role()`

~~As highlighted in the Reference PR, the naming still needs to be standardized (i.e. we're no longer just checking that a role exists but we're kinda checking for matches, so the method and the collections will have to reflect that)~~ (I think I tracked down all uses of 'exist' terminology and replaced it with 'match')

~~The versions of Reference may still need to be straightened out~~ (I think they're alright now)

~~Dependent on the Reference PR that implements the changes in the endpoint [(REFERENCE PR 202)](https://github.com/Health-Education-England/TIS-REFERENCE/pull/202)~~ (It's merged)

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PersonValidator.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PersonValidator.java
@@ -239,11 +239,11 @@ public class PersonValidator {
           .collect(Collectors.toList());
 
       Map<String, String> rolesMatch = referenceService.rolesMatch(roles, true);
-      List<String> correctedRoles = rolesMatch.entrySet().stream()
+      List<String> matchedRoles = rolesMatch.entrySet().stream()
           .map(entry -> entry.getValue().isEmpty() ? entry.getKey() : entry.getValue())
           .collect(Collectors.toList());
 
-      personDto.setRole(String.join(",", correctedRoles));
+      personDto.setRole(String.join(",", matchedRoles));
 
       for (Entry<String, String> roleMatch : rolesMatch.entrySet()) {
         if (roleMatch.getValue().isEmpty()) {
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PersonValidator.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PersonValidator.java
@@ -240,13 +240,13 @@ public class PersonValidator {
 
       Map<String, String> rolesMatch = referenceService.rolesMatch(roles, true);
       List<String> matchedRoles = rolesMatch.entrySet().stream()
-          .map(entry -> entry.getValue().isEmpty() ? entry.getKey() : entry.getValue())
+          .map(entry -> StringUtils.isEmpty(entry.getValue()) ? entry.getKey() : entry.getValue())
           .collect(Collectors.toList());
 
       personDto.setRole(String.join(",", matchedRoles));
 
       for (Entry<String, String> roleMatch : rolesMatch.entrySet()) {
-        if (roleMatch.getValue().isEmpty()) {
+        if (StringUtils.isEmpty(roleMatch.getValue())) {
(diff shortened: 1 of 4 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PersonValidator.java
```diff
@@ -238,14 +238,17 @@ public void validateForBulk(List<PersonDTO> personDtos) {
           .map(String::trim)
           .collect(Collectors.toList());
 
-      personDto.setRole(String.join(",", roles));
+      Map<String, String> rolesMatch = referenceService.rolesMatch(roles, true);
+      List<String> matchedRoles = rolesMatch.entrySet().stream()
+          .map(entry -> entry.getValue().isEmpty() ? entry.getKey() : entry.getValue())
```
JosephKelly, line 243: Looks like there's a potential for an NPE here.  You could use `StringUtils.isEmpty()`.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PersonValidator.java
```diff
@@ -238,14 +238,17 @@ public void validateForBulk(List<PersonDTO> personDtos) {
...
+          .collect(Collectors.toList());
 
-      Map<String, Boolean> rolesExist = referenceService.rolesExist(roles, true);
+      personDto.setRole(String.join(",", matchedRoles));
 
-      for (Entry<String, Boolean> roleExists : rolesExist.entrySet()) {
-        if (!roleExists.getValue()) {
+      for (Entry<String, String> roleMatch : rolesMatch.entrySet()) {
+        if (roleMatch.getValue().isEmpty()) {
```
JosephKelly, line 249: As above for NPE.

JosephKelly (changes requested):
If you wanted to test for it, you could add another entry with a null value to the map  in `PersonValidatorTest.bulkShouldGetErrorWhenRoleNotExists()#L210`

ReubenRobertsHEE (approved):

hihilary:
For `referenceService.rolesMatch`, I personally prefer a null value to a empty string when the role is not found. But this is subjective.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PersonValidator.java
Marcellofabbri: I _could_ guard it against NPEs to be extra careful. A null value can only exist if it's a case-insensitive match to a null key, which can't exist in a TreeMap, so it'd break in Reference first, either when trying to build the TreeMap, or when looking for matches for null strings. Also it would imply that the Role cell in the excel file has been left null, which would make the whole file be rejected before processing it at all.
But guarding it against NPEs is good anyway, in case the software around it changes in the future, so I've changed these lines to use StringUtils.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PersonValidator.java
Marcellofabbri: Used StringUtils like above

Marcellofabbri:
> If you wanted to test for it, you could add another entry with a null value to the map in `PersonValidatorTest.bulkShouldGetErrorWhenRoleNotExists()#L210`

Not sure this would make sense though, as the keys in the Map are nothing more than the substrings extrapolated from the multivalue Role field of the Person row. So the Map returned would never have an unexpected third entry if the extractable substrings are only two.

JosephKelly (approved):

PR #762, opened 2021-11-10
Fix/allow any medical school tis21 1635

JosephKelly:


Marcellofabbri (approved):
Seems to be working for me. It accepted whatever qualification I threw at it and wrote it in the Qualification table.

PR #769, opened 2021-11-22
feat: add validation to email to make unique

hihilary:
[TIS21_2284](https://hee-tis.atlassian.net/browse/TIS21-2284)

When updating the person record, the current one needs to be excluded for validation.

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/ContactDetailsValidator.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/ContactDetailsValidator.java
@@ -135,8 +135,9 @@ public class ContactDetailsValidator {
 
   private List<FieldError> checkEmail(ContactDetailsDTO dto) {
     List<FieldError> fieldErrors = new ArrayList<>();
-    validateEmailFormat("email", dto.getEmail(), fieldErrors);
-    validateEmailUniqueness("email", dto, fieldErrors);
+    String fieldName = "email";
+    validateEmailFormat(fieldName, dto.getEmail(), fieldErrors);
+    validateEmailUniqueness(fieldName, dto, fieldErrors);
     return fieldErrors;
   }
 
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/ContactDetailsValidator.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/ContactDetailsValidator.java
@@ -4,9 +4,13 @@ import com.transformuk.hee.tis.reference.api.dto.TitleDTO;
 import com.transformuk.hee.tis.reference.client.impl.ReferenceServiceImpl;
 import com.transformuk.hee.tis.tcs.api.dto.ContactDetailsDTO;
 import com.transformuk.hee.tis.tcs.service.model.ContactDetails;
+import com.transformuk.hee.tis.tcs.service.repository.ContactDetailsRepository;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
+import org.springframework.core.MethodParameter;
 import org.springframework.stereotype.Component;
 import org.springframework.validation.BeanPropertyBindingResult;
 import org.springframework.validation.FieldError;
@@ -35,9 +39,12 @@ public class ContactDetailsValidator {
       "Only numerical values allowed for %s, no special characters, with the exception of plus, minus and spaces.";
 
   private final ReferenceServiceImpl referenceService;
+  private final ContactDetailsRepository contactDetailsRepository;
 
-  public ContactDetailsValidator(ReferenceServiceImpl referenceService) {
+  public ContactDetailsValidator(ReferenceServiceImpl referenceService,
+      ContactDetailsRepository contactDetailsRepository) {
     this.referenceService = referenceService;
+    this.contactDetailsRepository = contactDetailsRepository;
   }
 
   /**
@@ -51,12 +58,16 @@ public class ContactDetailsValidator {
     final boolean currentOnly = false;
     List<FieldError> fieldErrors = new ArrayList<>();
     fieldErrors.addAll(checkTitle(dto, currentOnly));
+    fieldErrors.addAll(checkEmail(dto));
 
     if (!fieldErrors.isEmpty()) {
       BeanPropertyBindingResult bindingResult =
           new BeanPropertyBindingResult(dto, CONTACT_DETAILS_DTO_NAME);
       fieldErrors.forEach(bindingResult::addError);
-      throw new MethodArgumentNotValidException(null, bindingResult);
+
+      Method method = this.getClass().getMethods()[0];
+      MethodParameter methodParameter = new MethodParameter(method, 0);
+      throw new MethodArgumentNotValidException(methodParameter, bindingResult);
     }
   }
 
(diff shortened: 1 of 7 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/ContactDetailsValidator.java
```diff
@@ -124,14 +135,42 @@ private void validatePhoneNumber(String fieldName, String value, List<FieldError
...
-    validateEmail("email", dto.getEmail(), fieldErrors);
+    String fieldName = "email";
+    validateEmailFormat(fieldName, dto.getEmail(), fieldErrors);
+    validateEmailUniqueness(fieldName, dto, fieldErrors);
     return fieldErrors;
   }
 
-  private void validateEmail(String fieldName, String value, List<FieldError> fieldErrors) {
+  private void validateEmailFormat(String fieldName, String value, List<FieldError> fieldErrors) {
     if (value != null && !value.matches(REGEX_EMAIL)) {
```
ReubenRobertsHEE, line 145: Given that there's a TODO around improving this  on the REGEX_EMAIL string definition, I wonder if would be worth using the @Email validation, instead of a custom regex? There's a bit of a clunky implementation of this sort of idea here: https://github.com/Health-Education-England/TIS-GENERIC-UPLOAD/blob/19b4b0db688b0080b3233c1e208e3ded24119c18/generic-upload-service/src/main/java/com/transformuk/hee/tis/genericupload/service/service/PersonTransformerService.java#L899 . That would at least mean that the same rules are used in generic-upload and TCS.

I don't think its needed for this ticket, but it might be worth a thought.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/ContactDetailsValidator.java
```diff
@@ -51,12 +58,16 @@ public void validate(ContactDetailsDTO dto) throws MethodArgumentNotValidExcepti
     final boolean currentOnly = false;
     List<FieldError> fieldErrors = new ArrayList<>();
     fieldErrors.addAll(checkTitle(dto, currentOnly));
+    fieldErrors.addAll(checkEmail(dto));
```
ReubenRobertsHEE, line 61: THe javadoc above this method seems a bit out of date now.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/ContactDetailsValidator.java
```diff
@@ -51,12 +58,16 @@ public void validate(ContactDetailsDTO dto) throws MethodArgumentNotValidExcepti
...
 
     if (!fieldErrors.isEmpty()) {
       BeanPropertyBindingResult bindingResult =
           new BeanPropertyBindingResult(dto, CONTACT_DETAILS_DTO_NAME);
       fieldErrors.forEach(bindingResult::addError);
-      throw new MethodArgumentNotValidException(null, bindingResult);
+
+      Method method = this.getClass().getMethods()[0];
```
ReubenRobertsHEE, line 68: Isn't this a little brittle? I'm not an expert with using reflection, but it seems a bit of a gotcha to reference the ContactDetailsValidator through its position in the public methods array. I could be wrong, so please disregard if I've missed something.

ReubenRobertsHEE: For example, I might have thought this.getClass().getMethod("ContactDetailsValidator") might have been a bit clearer?

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/ContactDetailsValidator.java
```diff
@@ -124,14 +135,42 @@ private void validatePhoneNumber(String fieldName, String value, List<FieldError
...
+      });
+    }
+    int existingSize = existingContactDetails.size();
+    if (existingSize > 0) {
       fieldErrors.add(new FieldError(CONTACT_DETAILS_DTO_NAME, fieldName,
-          String.format(REGEX_EMAIL_ERROR, fieldName)));
+          String.format(
+              "email %s is not unique, there %s currently %d %s with this email (Person ID: %s)",
+              dto.getEmail(), existingSize > 1 ? "are" : "is", existingSize,
+              existingSize > 1 ? "emails" : "email",
```
ReubenRobertsHEE, line 170: "persons" : "person" 
is maybe clearer?

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/ContactDetailsValidator.java
hihilary: The @Email annotation is used in [ContactDetailsDTO](https://github.com/Health-Education-England/TIS-TCS/blob/1f4de2d0ec47bb58125b715b1c2f951ca82989c7/tcs-api/src/main/java/com/transformuk/hee/tis/tcs/api/dto/ContactDetailsDTO.java#L76). But in hibernate-validator-6.x, this annotation is deprecated (https://docs.jboss.org/hibernate/stable/validator/api/org/hibernate/validator/constraints/Email.html), while generic upload service uses hibernate-validator-5.1.3. 

I've no idea why we need the custom regex. It really worths a check.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/ContactDetailsValidator.java
hihilary: I found if we set the first parameter of `MethodArgumentNotValidException` to null, the `getMessage()` of the Exception class will throw an NPE in the console. So I copy-pasted those reflection codes from another validator.
Thank you for raising this problem. I think the method used here should be the current `validate` method (the 2nd one). Besides, `getMethod` throws an NoSuchMethodException, I have nothing to deal with it, so just throw it to the framework.

ReubenRobertsHEE (approved):
LGTM

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/ContactDetailsValidator.java
ReubenRobertsHEE: Works for me! :)

heetis-shivanirana:
@hihilary Have we also updated the e2e test for this functionality to test the the below AC:

Given validation is added on TIS 
When I update an existing trainee record’s email address on TIS front end (admins UI)
And provide an email address that already exist against another trainee record
Then I should be prevented from doing so via a clear message to the user
(c.f. similar to when updating the one of the registration numbers with a non-unique value)

Given validation is added on TIS 
When I update an existing trainee record’s other Personal details where the email address historically exist as a duplicate to another TIS trainee record
Then I should be prevented from doing so via a clear message until I correct the email address to be Unique to the trainee

Given validation is added on TIS 
When I use the Bulk People Upload with the Email address field populated on the template
And the Email address matches to a different trainee record on TIS (different registration number/different TIS person ID)
Then I should receive an error in the downloadable error report for that row

Given validation is added on TIS 
When I use the Bulk People Update with the Email address field populated on the template
And the Email address matches to a different trainee record on TIS (different registration number/different TIS person ID)
Then I should receive an error in the downloadable error report for that row
(c.f. similar to when updating the one of the registration numbers with a non-unique value)

hihilary:
@heetis-shivanirana I will check the e2e tests of the TIS front end. 
But for bulk upload, we don't have existing e2e tests set up for People Create/Update.

heetis-shivanirana:
Ok Great. Can you also please create a story to add the upload create/update e2e tests and place that in the backlog.

hihilary:
@heetis-shivanirana The assessment bulk upload e2e tests were added by me 2 sprints ago, before I added them there were only placement bulk upload e2e tests there, and half of them are tagged with TODO (xit). I don't know why they were excluded. Have you got any ideas? 
Maybe we need to sort out the excluded ones first?

heetis-shivanirana:
@hihilary I have asked to create the ticket for people bulk upload create/update that you said don't exist so that we have a story to implement them in future sprints.

PR #772, opened 2021-11-25
Revert/jackson update temporarily

JosephKelly:
Will reapply once I've taken a little extra time to add a test as well as get the service up to the same level


Judge40 (approved):
😞

PR #777, opened 2021-12-10
feat: add EsrEvent in PlacementDetailDTO

hihilary:
We've already had a `EsrEvent` set in PlacementSummaryDTO which used by the UI to show ESR status on Person-Placements list.
This ticket is to add the same `EsrEvent` set in PlacementDetailDTO for Placement Detail pages.

TIS21-174

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PlacementServiceImpl.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PlacementServiceImpl.java
@@ -801,10 +801,12 @@ public class PlacementServiceImpl implements PlacementService {
   }
 
   protected void populateEsrEventForPlacementDetail(PlacementDetailsDTO placementDetailsDto) {
-    Set<PlacementEsrEvent> esrEvents = placementEsrEventRepository.findPlacementEsrEventByPlacementIdIn(
-        Collections.singletonList(placementDetailsDto.getId()));
+    Set<PlacementEsrEvent> esrEvents = placementEsrEventRepository.
+        findPlacementEsrEventByPlacementIdIn(Collections.singletonList(
+            placementDetailsDto.getId()));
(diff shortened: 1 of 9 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PlacementServiceImpl.java
```diff
@@ -799,6 +800,16 @@ public PlacementDTO closePlacement(final Long placementId) {
...
   }
 
+  protected void populateEsrEventsForPlacementDetail(PlacementDetailsDTO placementDetailsDto) {
+    Set<PlacementEsrEvent> esrEvents = placementEsrEventRepository.
+        findPlacementEsrEventByPlacementIdIn(Collections.singletonList(
+            placementDetailsDto.getId()));
+    Set<PlacementEsrEventDto> esrEventDtos = new HashSet<>();
+    esrEvents.forEach(e -> esrEventDtos.add(
+        placementEsrEventDtoMapper.placementEsrEvenToPlacementEsrEventDto(e)));
+    placementDetailsDto.setEsrEvents(esrEventDtos);
```
JosephKelly, line 810: There are now a couple of places where this is done so I think this would be simpler if you follow the pattern of other mappers and define a `Set<?> placementEsrEventsToPlacementEsrEventDtos(Set<?>)`

JosephKelly (commented):
The change itself looks good.

Would now / a quick followup be a good time to fix the name of this existing method?
`placementEsrEventToPlacementEsrEventDto` is missing a `t`. It is:
`placementEsrEvenToPlacementEsrEventDto`

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PlacementServiceImpl.java
hihilary: Fixed, thanks for pointing it out.

JosephKelly (approved):

PR #786, opened 2021-12-16
Fix/check npn update is allowed

Marcellofabbri:
The current validation for NPNs of an updating Placement would only work for placements being updated normally via a Placement's page (which is prevented by the FE anyway).

When Generic Upload tries to update placements, what is being used in a call to the api is:
- the original placementDetailsDto as currently stored in the database
- some other fields, including the **new postId** related to the new NPN entered in the excel file
[[See here]](https://github.com/Health-Education-England/TIS-GENERIC-UPLOAD/blob/0669a2fd9dfc8d4572c29b1b054889b8a410d428/generic-upload-service/src/main/java/com/transformuk/hee/tis/genericupload/service/service/PlacementUpdateTransformerService.java#L129)

That's why the PlacementDetailsDto used in a PUT call to TCS at [[this line]](https://github.com/Health-Education-England/TIS-GENERIC-UPLOAD/blob/0669a2fd9dfc8d4572c29b1b054889b8a410d428/generic-upload-service/src/main/java/com/transformuk/hee/tis/genericupload/service/service/PlacementUpdateTransformerService.java#L232) has a **new postId** but an old **NPN**.

Structuring the validation like in this PR makes the validation work regardless whether the update is from bulk-upload or other FE components. The logic is based on using only the PlacementDetailsDto's `id` and `postId`.

- find the **old NPN** by finding the PlacementDetailsDto in the database and extracting it's NPN
- find the **new NPN** by finding a Post in the DB via the `PlacementDetailsDto.postId` (which is new), and extracting its NPN

This way we don't rely on the NPN carried by the PlacementDetailsDto itself, which is the old one when coming from bulk-upload, but the new one when coming from the main FE component.


Judge40 (changes requested):
We shouldn't be deleting any migration files

Marcellofabbri:
Those two deletions were not meant to be swept in with a push, they were meant to stay local.
I'll open a new PR, otherwise I'd need to revert the changes and mess with the commit history.

PR #787, opened 2021-12-16
Fix/check npn update is allowed tweak

Marcellofabbri:
The current validation for NPNs of an updating Placement would only work for placements being updated normally via a Placement's page (which is prevented by the FE anyway).

When Generic Upload tries to update placements, what is being used in a call to the api is:
- the original placementDetailsDto as currently stored in the database
- some other fields, including the **new postId** related to the new NPN entered in the excel file
[[See here]](https://github.com/Health-Education-England/TIS-GENERIC-UPLOAD/blob/0669a2fd9dfc8d4572c29b1b054889b8a410d428/generic-upload-service/src/main/java/com/transformuk/hee/tis/genericupload/service/service/PlacementUpdateTransformerService.java#L129)

That's why the PlacementDetailsDto used in a PUT call to TCS at [[this line]](https://github.com/Health-Education-England/TIS-GENERIC-UPLOAD/blob/0669a2fd9dfc8d4572c29b1b054889b8a410d428/generic-upload-service/src/main/java/com/transformuk/hee/tis/genericupload/service/service/PlacementUpdateTransformerService.java#L232) has a **new postId** but an old **NPN**.

Structuring the validation like in this PR makes the validation work regardless whether the update is from bulk-upload or other FE components. The logic is based on using only the PlacementDetailsDto's `id` and `postId`.

- find the **old NPN** by finding the PlacementDetailsDto in the database and extracting it's NPN
- find the **new NPN** by finding a Post in the DB via the `PlacementDetailsDto.postId` (which is new), and extracting its NPN

This way we don't rely on the NPN carried by the PlacementDetailsDto itself, which is the old one when coming from bulk-upload, but the new one when coming from the main FE component.

### DIFF
--- a/tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/validation/PlacementValidatorTest.java
+++ b/tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/validation/PlacementValidatorTest.java
@@ -56,6 +56,7 @@ public class PlacementValidatorTest {
   private static final String DEFAULT_PLACEMENT_TYPE = "OOPT";
   private static final BigDecimal DEFAULT_PLACEMENT_WHOLE_TIME_EQUIVALENT = new BigDecimal(1);
   private static final String DEFAULT_NATIONAL_POST_NUMBER = "NATIONAL_POST_NUMBER";
+  private static final String UPDATED_NATIONAL_POST_NUMBER = "UPDATED_NATIONAL_POST_NUMBER";
   private PlacementDetailsDTO placementDTO;
 
   @Mock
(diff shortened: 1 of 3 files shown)


hihilary:
Have gone through related codes in Generic upload, the `postId` is populated from the new `nationalPostNumber`. 
In tcs, the `nationalPostNumber` does not exist in `PlacementDetails`, which means when converting the `PlacementDetailsDTO` to `PlacementDetails`, the `nationalPostNumber` field is lost. So I'd guess the `nationalPostNumber` field of `PlacementDetailsDTO` is redundent. 
And for this ticket, yeah, I think getting the `nationalPostNumber` from Post could be a better solution. But I don't agree with the newly added tests, because all the fields of the DTO from bulk upload should have the same meaning as from the UI. To follow boyscott rule, the npn should be set to correct value in bulk upload.

Marcellofabbri:
I tried not to touch other services to keep the changes contained within TCS, but we could easily add this one line in Generic Upload:
`dbPlacementDetailsDTO.setNationalPostNumber(postDTO.getNationalPostNumber());` straight after [[this line]](https://github.com/Health-Education-England/TIS-GENERIC-UPLOAD/blob/0669a2fd9dfc8d4572c29b1b054889b8a410d428/generic-upload-service/src/main/java/com/transformuk/hee/tis/genericupload/service/service/PlacementUpdateTransformerService.java#L225).

And the issue should be resolved with no need to change TCS any further (this PR too would be superfluous because TCS as it is now would receive both the updated `postId` and the updated `nationalPostNumber`).

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/validation/PlacementValidatorTest.java
```diff
@@ -305,27 +306,57 @@ public void multipleCommentsShouldThrowValidationException() throws ValidationEx
   }
 
   @Test(expected = ValidationException.class)
-  public void validateNpnForPlacementExportedToEsr() throws ValidationException {
-    // arrange old Placement already present in DB
-    Post dbPost = new Post();
-    dbPost.setId(DEFAULT_POST);
-    dbPost.setNationalPostNumber(DEFAULT_NATIONAL_POST_NUMBER);
+  public void validateNpnForPlacementExportedToEsrWhenUpdatingViaBulkUpload()
```
hihilary, line 309: I don't think the tests need to be modified particularly for BulkUpload.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/validation/PlacementValidatorTest.java
```diff
@@ -305,27 +306,57 @@ public void multipleCommentsShouldThrowValidationException() throws ValidationEx
...
+    given(ownerProjection.getNationalPostNumber()).willReturn(UPDATED_NATIONAL_POST_NUMBER);
+    given(postRepository.findPostById(456L)).willReturn(ownerProjection);
+
+    // act
+    placementValidator.validate(placementDTO);
+  }
 
-    // arrange new PlacementDetailsDto being passed in by the user
+  @Test(expected = ValidationException.class)
+  public void validateNpnForPlacementExportedToEsrWhenUpdatingViaPlacementPage()
```
hihilary, line 337: The same as the above one.

PR #791, opened 2021-12-23
fix: NPE in isEligibleForCurrentTraineeWteChangeNotification method

hihilary:
When updating the `wte` in a placement, backend calls a `isEligibleForCurrentTraineeWteChangeNotification` method which compares if the current `wte` and the new `wte` are the same. If the current `wte` is null, it throws a NPE, which results in a 500 error on the UI.

TIS21-2441

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PlacementServiceImpl.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PlacementServiceImpl.java
@@ -1010,11 +1010,11 @@ public class PlacementServiceImpl implements PlacementService {
     LocalDate currentDateFrom = latestApprovedLog.getDateFrom();
 
     return
-        ((updatedPlacementDetails.getWholeTimeEquivalent() != null
-            && (currentPlacement.getPlacementWholeTimeEquivalent() == null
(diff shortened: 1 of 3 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PlacementServiceImpl.java
```diff
@@ -1010,10 +1010,11 @@ public boolean isEligibleForCurrentTraineeWteChangeNotification(final Placement
     LocalDate currentDateFrom = latestApprovedLog.getDateFrom();
 
     return
-        ((updatedPlacementDetails.getWholeTimeEquivalent() != null && updatedPlacementDetails
-            .getWholeTimeEquivalent()
-            .compareTo(currentPlacement.getPlacementWholeTimeEquivalent()) != 0)) &&
-            ((currentDateFrom != null && currentDateFrom
+        (updatedPlacementDetails.getWholeTimeEquivalent() != null && (
+            currentPlacement.getPlacementWholeTimeEquivalent() == null
```
hihilary, line 1014: This condition of checking null value is newly added, and the rest modified lines are for formating.

Marcellofabbri (approved):
Looks fine

PR #798, opened 2021-12-24
Feat: ProgrammeMembership to CurriculumMembership 

ReubenRobertsHEE:
The intention of this PR is to establish a new CurriculumMembership table as a copy of the existing ProgrammeMembership table, and to route all relevant data persistence (CRUD) to both of these tables in parallel. 

TCS Service ProgrammeMembership functionality has been rewritten to refer to the new table via the mapper and amended ProgrammeMembershipServiceImpl. The Person object is also amended to refer to the new CurriculumMembership. These changes will be required for (*) below, which is also part of this ticket. 

Note that the ProgrammeMembershipDTO is unchanged, and will remain unchanged until much later in the epic: this should mean no modifications are needed for functionality using this object (e.g. events emitted to downstream Reval interfaces).

TSS Sync should continue to function with the DMS events from the original ProgrammeMembership table; the flow of these events should be unaffected by this PR. 

Next steps:
1. Modify TSS Sync to rather use DMS events from the new CurriculumMembership table.
2. Modify TIS-TCS to remove parallel CRUD persistence so that ongoing data updates are only applied to the CurriculumMembership table (*). 
3. At that point, nothing should be using the ProgrammeMembership table, which can then be refactored into the new normalised data model (https://hee-tis.atlassian.net/browse/TIS21-2421)

TIS21-2420


ReubenRobertsHEE:
The 6% code duplication is unavoidable: the new CurriculumMembership class is intended to be an exact copy of the existing ProgrammeMembership class, so the duplication is 'by design'. Later in the epic (TIS21-2421) we will be refactoring the defunct ProgrammeMembership class to refer to the refactored ProgrammeMembership table, which will resolve the duplication.

ReubenRobertsHEE:
This is not ready to deploy: currently getting a 'You are acting on stale data, please refresh' concurrency failure on the _second_ edit of a person's programme memberships.

ReubenRobertsHEE:
This PR is a bit messy: a clean version is now here: https://github.com/Health-Education-England/TIS-TCS/pull/802

PR #802, opened 2021-12-30
feat: ProgrammeMembership to CurriculumMembership

ReubenRobertsHEE:
The intention of this PR is to establish a new CurriculumMembership table as a copy of the existing ProgrammeMembership table, and to route all relevant data persistence (CRUD) to both of these tables in parallel.

TCS Service ProgrammeMembership functionality has been rewritten to refer to the new table via the mapper and amended ProgrammeMembershipServiceImpl. The Person object is also amended to refer to the new CurriculumMembership. These changes will be required for (*) below, which is also part of this ticket.

Note that the ProgrammeMembershipDTO is unchanged, and will remain unchanged until much later in the epic: this should mean no modifications are needed for functionality using this object (e.g. events emitted to downstream Reval interfaces).

TSS Sync should continue to function with the DMS events from the original ProgrammeMembership table; the flow of these events should be unaffected by this PR.

Dependent on: https://hee-tis.atlassian.net/browse/TIS21-2381 being merged first (to avoid having to do any data clean-ups in two places at once).

Next steps:

1. Modify TSS Sync to rather use DMS events from the new CurriculumMembership table; modify TIS Sync to use CurriculumMembership table; modify NDW-ETL to use CurriculumMembership table.
2. Modify TIS-TCS to remove parallel CRUD persistence so that ongoing data updates are only applied to the CurriculumMembership table (*).
3. At that point, nothing should be using the ProgrammeMembership table, which can then be refactored into the new normalised data model (https://hee-tis.atlassian.net/browse/TIS21-2421)
TIS21-2420


ReubenRobertsHEE:
The code duplication is unavoidable I think: the new CurriculumMembership class is intended to be an exact copy of the existing ProgrammeMembership class, so the duplication is 'by design'. Later in the epic (TIS21-2421) we will be refactoring the defunct ProgrammeMembership class to refer to the refactored ProgrammeMembership table, which will resolve the duplication.

The code coverage on CurriculumMembership is a bit weird: locally, I get 100% method + 98% line coverage, so I'm not sure why Sonar thinks its 0%.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ProgrammeMembershipServiceImpl.java
```diff
@@ -350,4 +389,39 @@ private void updatePersonWhenStatusIsStale(Long personId) {
...
+   * Set a ProgrammeMembership AmendedDate from its database record.
+   *
+   * <p>The amendedDate of the programmeMembershipDto which is being saved is based on the entry
+   * in the CurriculumMembership table, which will have a slightly different AmendedDate value
+   * from the record in the ProgrammeMembership table. Since this field is used for @Version
+   * optimistic locking, we need the actual value from the ProgrammeMembership table otherwise we
+   * will get 'You are acting on stale data, please refresh' concurrency failures.
+   *
+   * <p>Caution: by overwriting the AmendedDate, it is possible to handle concurrent updates to the
+   * same ProgrammeMembership incorrectly: for exampel, if record A is updated to A' and this is
```
hihilary, line 403: typo 'exampel'

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ProgrammeMembershipServiceImpl.java
```diff
@@ -350,4 +389,39 @@ private void updatePersonWhenStatusIsStale(Long personId) {
...
+   * B needs to be based on A', otherwise the A' update is lost. To avoid this, apply the
+   * ProgrammeMembership record Save after the parallel CurriculumMembership Save in the same
+   * code block: in the event of stale data (the B update in the example above) the
+   * CurriculumMembership Save will fail and throw an exception before the incorrect
+   * ProgrammeMembership Save is applied.
+   *
+   * @param pm the ProgrammeMembership to set the amended date for
+   * @return the updated ProgrammeMembership
+   */
+  private ProgrammeMembership setProgrammeMembershipAmendedDateFromDbRecord(
```
hihilary, line 414: I found the UI always uses the lastModifiedDate which is received from the backend for `amendedDate` in `ProgrammeMembershipDTO`, so I'm a little wondering if the `amendedDate` needs to be set again here.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ProgrammeMembershipServiceImpl.java
```diff
@@ -93,26 +112,36 @@ public ProgrammeMembershipDTO save(ProgrammeMembershipDTO programmeMembershipDTO
...
    * Save a list of programmeMembership.
    *
-   * @param programmeMembershipDTO the list of entities to save
+   * @param programmeMembershipDto the list of entities to save
    * @return the list of persisted entities
    */
   @Override
-  public List<ProgrammeMembershipDTO> save(List<ProgrammeMembershipDTO> programmeMembershipDTO) {
-    log.debug("Request to save ProgrammeMembership : {}", programmeMembershipDTO);
+  public List<ProgrammeMembershipDTO> save(List<ProgrammeMembershipDTO> programmeMembershipDto) {
```
hihilary, line 119: I'm thinking if we might need to add @Transactional annotation for save/delete.

hihilary:
Under tcs-persistence/.../resource/queries folder, there're lots of sql queries used by TIS-SYNC and elasticsearch, etc. The `ProgrammeMembership` needs to be switched to `CurriculumMembership` too.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ProgrammeMembershipServiceImpl.java
ReubenRobertsHEE: It's declared at the class level, so it would apply here as well :)

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ProgrammeMembershipServiceImpl.java
ReubenRobertsHEE: If I understand the comment correctly, you're wondering if we need to bother setting the amendedDate, since it should be correctly passed from the UI?
The issue is that when we save() a record, it is saved to both the ProgrammeMembership and CurriculumMembership tables - but the amendedDate field in the database for each record is populated with a timestamp, which differs between the two saved records by some milliseconds. Since we then only load one version of the record (from CurriculumMembership) we get an amendedDate value from that record - and then when we later try to save an updated version of it as a ProgrammeMembership record (to keep them updated in parallel), the database rejects it because the amendedDate does not match the value it has for that record in the ProgrammeMembership table. This is part of the optimistic-locking check to make sure you are not saving an outdated version of a record over a newer version. The code here is a little bit hacky and I'm not 100% happy about it, but I can't think of an elegant alternative: loading and doing all the processing to both versions of the record (from the CurriculumMembership and ProgrammeMembership tables) seemed likely to be clumsy and verbose, but I am very open to alternatives to the method I've implemented, which I would certainly reject myself if the parallel-save code was intended to be more than temporary.

ReubenRobertsHEE:
> Under tcs-persistence/.../resource/queries folder, there're lots of sql queries used by TIS-SYNC and elasticsearch, etc. The `ProgrammeMembership` needs to be switched to `CurriculumMembership` too.

Thanks - great catch!

hihilary:
LGTM

hihilary (dismissed):

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ProgrammeMembershipServiceImpl.java
```diff
@@ -350,4 +389,39 @@ private void updatePersonWhenStatusIsStale(Long personId) {
       personRepository.save(person);
     }
   }
+
```
ReubenRobertsHEE, line 392: @JosephKelly / @Judge40 - I wonder if I could get your opinion on this? The rest of the PR is very standard stuff, but there is the little quirk below in relation to saving a record into two tables. It could be that we can live with this, given the parallel-saves are intended to be temporary while we finish the ProgrammeMembership refactoring, but I'm wondering if there might be a more elegant way of dealing with the issue. Happy to chat about it anytime!

JosephKelly:
> The rest of the PR is very standard stuff, but there is the little quirk below in relation to saving a record into two tables. It could be that we can live with this, given the parallel-saves are intended to be temporary while we finish the ProgrammeMembership refactoring, but I'm wondering if there might be a more elegant way of dealing with the issue. Happy to chat about it anytime!

A chat might be the best way to go.  I am wondering if you'd discussed/considered separating the writes, whether it would just complicate it to have the legacy structure modified by subsequent CRUD operations or just not update it at all!

hihilary (approved):

PR #804, opened 2022-01-05
Feat/get specialties by specialty type tis21 2460

Marcellofabbri:
In **Posts**, under **ALLOCATION**, the `Sub specialties` field currently displays all specialties instead of showing only specialties whose SpecialtyType is `SUB_SPECIALTY`.

image

The `<tis-dropdown>` element from the frontend is hitting the `getAllSpecialties()` endpoint in TCS. It can already retrieve all specialties according to Status (=CURRENT), but can't yet retrieve all specialties according to SpecialtyType for a number of reasons:
- SpecialtyType in Specialty is lazily instantiated
- Even when made transactional, it still can't cope because it's not a simple field like `Status`, but a `Set<SpecialtyType>`

By transforming each value of the SpecialtyType columnFilter in its own spec that uses the CriteriaBuilder method `isMember` instead of `in`, when given any SpecialtyType in the ColumnFilters, the endpoint can now handle searches by SpecialtyType as well.

In this PR I'm only looking to fix the `getAllSpecialties()` endpoint, which looked like it could handle SpecialtyType columnFilters but actually couldn't. In other PRs I was going to address backend validations.

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/SpecialtyServiceImpl.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/SpecialtyServiceImpl.java
@@ -121,7 +121,7 @@ public class SpecialtyServiceImpl implements SpecialtyService {
     if (columnFilters != null && !columnFilters.isEmpty()) {
       columnFilters.forEach(cf -> {
         if (StringUtils.equals(cf.getName(), "specialtyTypes")) {
-          List<Object> specialtyTypesValues = cf.getValues().stream()
+          List<SpecialtyType> specialtyTypesValues = cf.getValues().stream()
               .map(value -> SpecialtyType.valueOf((String) value))
               .collect(Collectors.toList());
 
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/SpecialtyServiceImpl.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/SpecialtyServiceImpl.java
@@ -125,7 +125,8 @@ public class SpecialtyServiceImpl implements SpecialtyService {
               .map(value -> SpecialtyType.valueOf((String) value))
               .collect(Collectors.toList());
 
-          specialtyTypesValues.forEach(specialtyTypeValue -> specs.add(isMember(cf.getName(), specialtyTypeValue)));
+          specialtyTypesValues.forEach(specialtyTypeValue ->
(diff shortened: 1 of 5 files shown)


hihilary:
Taking a look at the `PostSpecialty` DB table, I think the filter for the sub specialties should be `PostSpecialtyType` instead of `SpecialtyType`.

hihilary (changes requested):
Taking a look at the `PostSpecialty` DB table, I think the filter for the sub specialties should be `PostSpecialtyType` instead of `SpecialtyType`.

hihilary:
Sorry, i think I was wrong.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/SpecialtyServiceImpl.java
```diff
@@ -116,7 +119,18 @@ public SpecialtyDTO save(SpecialtyDTO specialtyDTO) {
     }
     //add the column filters criteria
     if (columnFilters != null && !columnFilters.isEmpty()) {
-      columnFilters.forEach(cf -> specs.add(in(cf.getName(), cf.getValues())));
+      columnFilters.forEach(cf -> {
+        if (StringUtils.equals(cf.getName(), "specialtyTypes")) {
+          List<Object> specialtyTypesValues = cf.getValues().stream()
```
hihilary, line 124: Use `SpecialtyType` instead of `Object`

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/SpecialtyResourceIntTest.java
```diff
@@ -438,6 +442,50 @@ public void shouldTextSearchAndFilterColumns() throws Exception {
...
+    // Get all the specialties that match at least the columnFilters. Should retrieve:
+    // specialty2 (with specialtyTypes SUB_SPECIALTY and PLACEMENT)
+    // specialty (with specialtyTypes SUB_SPECIALTY, PLACEMENT and CURRICULUM)
+    restSpecialtyMockMvc
+        .perform(get("/api/specialties?sort=id,desc&columnFilters=" + colFilters))
+        .andExpect(status().isOk())
+        .andExpect(jsonPath("$.[0].id").value(specialty2.getId()))
+        .andExpect(jsonPath("$.[1].id").value(specialty.getId()))
+        .andExpect(jsonPath("$.[2]").doesNotExist());
+    // and nothing else
```
hihilary, line 486: Please add more verifications to make sure the `specialtyTypes` is present.

hihilary (approved):

PR #808, opened 2022-01-10
Feat/add specialty validation for subspecialties tis21 2460

Marcellofabbri:
Addition of a validation on Specialties when creating or updating a Post.

When the `createPost()` and `updatePost()` endpoints are hit, the PostDto in question goes through validations concerning sites, grades, specialties etc.

When checking the validity of its specialties, the Post is inspected to make sure of the following:
- check if there's any `PostSpecialty` of PostSpecialtyType `SUB_SPECIALTY` (i.e. check if there's any Specialty that has been indicated as the Post's "subspecialty"
- if there is: check that the Specialty contained in the PostSpecialty is of type `SUB_SPECIALTY` (i.e. check that the Specialty indicated as the Post's "subspecialty" is itself, indeed, a Specialty of SpecialtyType `SUB_SPECIALTY`

I tried to uniform this validation in line with other validations, nevertheless, when trying to create/update a Post via Generic Upload, when Generic Upload receives a response from TCS with the exception, it triggers the writing of an unrelated message on the Excel file. I believe that Generic Upload will need to be amended slightly to capture the exception and parse it, in order to write the correct error on the Excel file (which I think it already does with other datatypes like Placements)

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PostValidator.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PostValidator.java
@@ -9,7 +9,9 @@ import com.transformuk.hee.tis.tcs.api.dto.PostSiteDTO;
 import com.transformuk.hee.tis.tcs.api.dto.PostSpecialtyDTO;
 import com.transformuk.hee.tis.tcs.api.dto.ProgrammeDTO;
 import com.transformuk.hee.tis.tcs.api.enumeration.PostSpecialtyType;
+import com.transformuk.hee.tis.tcs.api.enumeration.SpecialtyType;
 import com.transformuk.hee.tis.tcs.service.model.Post;
+import com.transformuk.hee.tis.tcs.service.model.Specialty;
 import com.transformuk.hee.tis.tcs.service.repository.PlacementRepository;
 import com.transformuk.hee.tis.tcs.service.repository.PostRepository;
 import com.transformuk.hee.tis.tcs.service.repository.ProgrammeRepository;
@@ -20,6 +22,7 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -185,12 +188,22 @@ public class PostValidator {
                 String.format("Specialty with id %d does not exist", ps.getSpecialty().getId())));
           } else if (PostSpecialtyType.PRIMARY.equals(ps.getPostSpecialtyType())) {
             ++noOfPrimarySpecialtyCount;
+          } else if (PostSpecialtyType.SUB_SPECIALTY.equals(ps.getPostSpecialtyType())) {
+            Optional<Specialty> specialty = specialtyRepository.findSpecialtyByIdEagerFetch(ps.getSpecialty().getId());
+            specialty.ifPresent(s -> checkSubspecialtyIsOfTypeSubspecialty(fieldErrors, s));
           }
         }
       }
       checkSpecialtyType(fieldErrors, noOfPrimarySpecialtyCount);
     }
     return fieldErrors;
+  }
+
+  private void checkSubspecialtyIsOfTypeSubspecialty(List<FieldError> fieldErrors, Specialty specialty) {
+    if (!specialty.getSpecialtyTypes().contains(SpecialtyType.SUB_SPECIALTY)) {
+      fieldErrors.add(new FieldError(POST_DTO_NAME, SPECIALTIES,
+          String.format("The specialty %s is not a %s", specialty.getName(), SpecialtyType.SUB_SPECIALTY)));
+    }
   }
 
   private void checkSpecialtyType(List<FieldError> fieldErrors, int noOfPrimarySpecialtyCount) {
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PostValidator.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PostValidator.java
@@ -189,8 +189,8 @@ public class PostValidator {
           } else if (PostSpecialtyType.PRIMARY.equals(ps.getPostSpecialtyType())) {
             ++noOfPrimarySpecialtyCount;
           } else if (PostSpecialtyType.SUB_SPECIALTY.equals(ps.getPostSpecialtyType())) {
-            Optional<Specialty> specialty = specialtyRepository.findSpecialtyByIdEagerFetch(ps.getSpecialty().getId());
-            specialty.ifPresent(s -> checkSubspecialtyIsOfTypeSubspecialty(fieldErrors, s));
+            specialtyRepository.findSpecialtyByIdEagerFetch(ps.getSpecialty().getId())
+                .ifPresent(s -> checkSubspecialtyIsOfTypeSubspecialty(fieldErrors, s));
           }
         }
       }
--- a/tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
+++ b/tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
@@ -497,9 +497,9 @@ public class PostResourceIntTest {
   public void shouldFailToCreatePostIfSubspecialtyIsNotASubspecialty() throws Exception {
     Optional<Post> postBeforeUpdate = postRepository.findById(1L);
 
-    // Attempt to update a Post that has a specialty of specialtyType PLACEMENT as a subspecialty
-    // (i.e. link Post and a Specialty that is SpecialtyType.SUB_SPECIALTY in a PostSpecialty in
-    // a PostSpecialty of PostSpecialtyType.PLACEMENT)
+    // Attempt to save a Post that has a specialty of specialtyType PLACEMENT as a subspecialty
+    // (i.e. link Post and a Specialty that is SpecialtyType.PLACEMENT in a PostSpecialty in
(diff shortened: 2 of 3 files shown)


tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
```diff
@@ -397,6 +399,164 @@ public void shouldAllowMMultipleOtherSpecialties() throws Exception {
...
+        .filter(ps -> ps.getPostSpecialtyType().equals(PostSpecialtyType.SUB_SPECIALTY))
+        .collect(Collectors.toList());
+
+    assertThat(addedSubspecialties.size()).isZero();
+    assertThat(postAfterUpdate).isEqualTo(post);
+  }
+
+  @Test
+  @Transactional
+  public void shouldAllowUpdateOfPostIfSubspecialtyIsNotASubspecialty() throws Exception {
```
ReubenRobertsHEE, line 452: Should this be 'shouldAllowUpdateOfPostIfSubspecialtyIsASubspecialty()'? Otherwise I am very confused! :)

ReubenRobertsHEE (changes requested):

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
```diff
@@ -397,6 +399,164 @@ public void shouldAllowMMultipleOtherSpecialties() throws Exception {
...
+    // the Post needs a Primary specialty as well, otherwise it wouldn't be possible to set a
+    // sub_specialty
+    post = createEntity();
+    PostSpecialty primaryPostSpecialty = createPostSpecialty(specialty, PostSpecialtyType.PRIMARY, post);
+    post.setSpecialties(new HashSet<>(Arrays.asList(primaryPostSpecialty)));
+    postRepository.saveAndFlush(post);
+
+    // Attempt to update a Post that has a specialty of specialtyType SUB_SPECIALTY as a subspecialty
+    // (i.e. link Post and a Specialty that is SpecialtyType.SUB_SPECIALTY in a PostSpecialty in
+    // a PostSpecialty of PostSpecialtyType.SUB_SPECIALTY)
```
ReubenRobertsHEE, line 463: Looks like a bit of word duplication 'in a PostSpecialty in a PostSpecialty' which makes this seem possibly a bit confusing.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
```diff
@@ -397,6 +399,164 @@ public void shouldAllowMMultipleOtherSpecialties() throws Exception {
...
+  }
+
+  @Test
+  @Transactional
+  public void shouldAllowCreationOfPostIfSubspecialtyIsASubspecialty() throws Exception {
+    int databaseSizeBeforeCreate = postRepository.findAll().size();
+
+    // Attempt to save a Post that has a specialty of specialtyType PLACEMENT as a subspecialty
+    // (i.e. link Post and a Specialty that is SpecialtyType.PLACEMENT in a PostSpecialty in
+    // a PostSpecialty of PostSpecialtyType.SUB_SPECIALTY)
```
ReubenRobertsHEE, line 537: This comment is identical to the one in the preceding test, which is testing the opposite condition.

ReubenRobertsHEE: (There also looks to be word duplication here and above - if the duplication is intentional, then maybe there's another way to phrase it, because the current form of words is a bit tricky to digest.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PostValidator.java
```diff
@@ -193,6 +198,15 @@ private void notExistsFieldErrors(List<FieldError> fieldErrors,
     return fieldErrors;
   }
 
+  private void checkSubspecialtyIsOfTypeSubspecialty(List<FieldError> fieldErrors,
```
ReubenRobertsHEE, line 201: Perhaps just personal preference, but I would name this 'checkSpecialtyIsOfTypeSubspecialty' since you could pass in any sort of specialty, not just subspecialties. The logic in the call on line 192 could make it clear that a subspecialty was being tested by renaming 's' as 'subspecialty' or similar).

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
```diff
@@ -397,6 +399,164 @@ public void shouldAllowMMultipleOtherSpecialties() throws Exception {
         .isEqualTo(PostSpecialtyType.OTHER);
   }
 
+  @Test
+  @Transactional
+  public void shouldFailToUpdatePostIfSubspecialtyIsNotASubspecialty() throws Exception {
```
ReubenRobertsHEE, line 404: Again, possibly just a point of clarity, but it might be worth trying to distinguish from the postSpecialtyType and the  specialtyType to make it a bit easier to understand the test, e.g. calling this 'shouldFailToUpdatePostIfSubspecialtyIsNotOfTypeSubspecialty'

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
ReubenRobertsHEE: (And as per comment above, possibly rename to something like ''shouldAllowUpdateOfPostIfSubspecialtyIsOfTypeSubspecialty()'

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
```diff
@@ -397,6 +399,164 @@ public void shouldAllowMMultipleOtherSpecialties() throws Exception {
...
+        .filter(ps -> ps.getPostSpecialtyType().equals(PostSpecialtyType.SUB_SPECIALTY))
+        .collect(Collectors.toList());
+
+    assertThat(addedSubspecialties.size()).isEqualTo(1);
+    assertThat(addedSubspecialties.get(0).getSpecialty()).isEqualTo(aSubspecialty);
+  }
+
+  @Test
+  @Transactional
+  public void shouldFailToCreatePostIfSubspecialtyIsNotASubspecialty() throws Exception {
```
ReubenRobertsHEE, line 498: Maybe rename to ...IsNotOfTypeSubspecialty' for clarity

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
```diff
@@ -397,6 +399,164 @@ public void shouldAllowMMultipleOtherSpecialties() throws Exception {
...
+        .andExpect(jsonPath("$.message").value("error.validation"))
+        .andExpect(jsonPath("$.fieldErrors[0].field").value("specialties"));
+
+    List<Post> postList = postRepository.findAll();
+    assertThat(postList).hasSize(databaseSizeBeforeCreate);
+  }
+
+  @Test
+  @Transactional
+  public void shouldAllowCreationOfPostIfSubspecialtyIsASubspecialty() throws Exception {
```
ReubenRobertsHEE, line 532: Maybe rename to '...IsOfTypeSubspecialty' for clarity

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
Marcellofabbri: Good suggestion 👍. I've added "...OfType..." to all 4 tests.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
Marcellofabbri: Ah yes, I copied this more than once, then went back to edit them, and in the end they got all messed up.
I've rewritten all four tests' comments, hopefully they're clearer now.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
Marcellofabbri: As above, I've rewritten all 4 comments.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PostValidator.java
Marcellofabbri: Agreed. The method takes in any specialty and checks whether it's of `specialtyType.SUB_SPECIALTY` or not. I've renamed it accordingly.
But the `s ->` on line 192 is indeed a parameter of that method, and can be all sorts of specialties, not just a subspecialty, so I renamed it to `specialty`

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
Marcellofabbri: I've renamend all 4 tests to look like this.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
Marcellofabbri: same as above

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
Marcellofabbri: same as above

ReubenRobertsHEE (approved):
LGTM

PR #812, opened 2022-01-17
Feat/overload get specialty by name method

Marcellofabbri:
Extra method added so that `tcsServiceImpl.getSpecialtyByName` can take two parameters, not just one.
The second parameter is the `SpecialtyType`, and it'll pass the value as a columnFilter to the endpoint it'll call.

Addition needed in order to retrieve all matching Specialties of type **SpecialtyType.SUB_SPECIALTY** from Generic Upload.
[Here's where it'll be used](https://github.com/Health-Education-England/TIS-GENERIC-UPLOAD/pull/156/files#diff-6256dd397dc350eb0e80fce68b23e6999bb856d18c77e356bc23b841fa588e23R279)

Only thing I wasn't sure about was whether to give the new method the same name and overload the existing method, or to specify a different name (e.g. `getSpecialtiesByNameAndSpecialtyType`)

TIS21-2460: Limit list of specialties to subspecialties

### DIFF
--- a/tcs-client/src/test/java/com/transformuk/hee/tis/tcs/client/service/impl/TcsServiceImplTest.java
+++ b/tcs-client/src/test/java/com/transformuk/hee/tis/tcs/client/service/impl/TcsServiceImplTest.java
@@ -1,17 +1,21 @@
 package com.transformuk.hee.tis.tcs.client.service.impl;
 
 import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.not;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
 
 import com.github.tomakehurst.wiremock.junit.WireMockClassRule;
 import com.google.common.collect.Maps;
 import com.transformuk.hee.tis.tcs.api.dto.AbsenceDTO;
+import com.transformuk.hee.tis.tcs.api.dto.CurriculumDTO;
 import com.transformuk.hee.tis.tcs.api.dto.SpecialtyDTO;
 import com.transformuk.hee.tis.tcs.api.enumeration.SpecialtyType;
 import java.time.LocalDate;
@@ -22,10 +26,9 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
-
-import com.transformuk.hee.tis.tcs.api.dto.CurriculumDTO;
 import org.apache.commons.codec.EncoderException;
 import org.apache.commons.codec.net.URLCodec;
+import org.assertj.core.util.Lists;
 import org.junit.Before;
 import org.junit.ClassRule;
 import org.junit.Rule;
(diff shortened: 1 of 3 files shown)


tcs-client/src/test/java/com/transformuk/hee/tis/tcs/client/service/impl/TcsServiceImplTest.java
```diff
@@ -240,6 +249,30 @@ public void getSpecialtyByIdShouldFindSpecialtyDto() {
...
+    doReturn(responseEntity).when(restTemplate).exchange(url, HttpMethod.GET, null,
+        new ParameterizedTypeReference<List<SpecialtyDTO>>() {});
+
+    List<SpecialtyDTO> result = testObj.getSpecialtyByName("specialtyName",
+        SpecialtyType.SUB_SPECIALTY);
+    assertThat("Unexpected result", result, is(expectedResult));
+    verify(restTemplate).exchange(url, HttpMethod.GET, null,
+        new ParameterizedTypeReference<List<SpecialtyDTO>>() {});
+  }
+
```
ReubenRobertsHEE, line 275: It would be nice to test the opposite condition, i.e. that it doesn't find specialties by name if the type doesn't match. Maybe something like:

```
@Test
  public void getSpecialtyByNameShouldNotFindDifferentTypeSpecialtyDto() throws EncoderException {
    SpecialtyDTO specialty = new SpecialtyDTO();
    specialty.setName("specialtyName");
    specialty.setSpecialtyTypes(new HashSet<>(
        Collections.singletonList(SpecialtyType.SUB_SPECIALTY)));
    specialty.setId(20L);

    String encodedParameters = new URLCodec().encode("{\"name\":[\"specialtyName\"],"
        + "\"status\":[\"CURRENT\"],\"specialtyTypes\":[\"PLACEMENT\"]}");
    String url = "http://localhost:9999/tcs/api/specialties?columnFilters=" + encodedParameters;

    List<SpecialtyDTO> expectedResult = Lists.emptyList();
    ResponseEntity responseEntity = new ResponseEntity(expectedResult, HttpStatus.OK);
    doReturn(responseEntity).when(restTemplate).exchange(url, HttpMethod.GET, null,
        new ParameterizedTypeReference<List<SpecialtyDTO>>() {});
    
    exceptionRule.expect(HttpClientErrorException.class);
    exceptionRule.expectMessage("404 Not Found");

    List<SpecialtyDTO> result = testObj.getSpecialtyByName("specialtyName",
        SpecialtyType.SUB_SPECIALTY);
    assertThat("Unexpected result", result, is(expectedResult));
    verify(restTemplate).exchange(url, HttpMethod.GET, null,
        new ParameterizedTypeReference<List<SpecialtyDTO>>() {});
  }
```

Otherwise it all looks good! :)

Marcellofabbri: This test passes but I think it's for the wrong reasons, which is `restTemplate.exchange(<URL for SUB_SPECIALTY> etc...)` not being stubbed. The url for retrieving PLACEMENT types of Specialties is formulated but it's never used in the test, which still only calls `testObj.getSpecialtyByName("specialtyName", SpecialtyType.SUB_SPECIALTY)`, and therefore it'll try to use `restTemplate.exchange(<URL for SUB_SPECIALTY> etc...)` instead of `restTemplate.exchange(<URL for PLACEMENT> etc...)`.

The actual functionality under test here would be that given a name and specialty type to be passed to the method, when the method is called, then it results in a REST GET call with such parameters. Everything happening afterwards is responsibility of `restTemplate` and of the endpoint, which is tested elsewhere. Also I wouldn't expect the call to raise an exception with a 404, since it's looking for matching dtos, so I'd expect a 200 with an empty list.

I added this test and tweaked it so that it stubs both scenarios (SUB_SPECIALTY and non-SUB_SPECIALTY), and verifies that the resulting REST call is consistent with the argument being passed to the method and isn't with the other argument passed to the method.

ReubenRobertsHEE: Ah, that's true - I was being hasty.

ReubenRobertsHEE (approved):

PR #815, opened 2022-01-17
chore: Update tcs-client version

Marcellofabbri:
Forgot to bump tcs-client version up last time I merged.

Before approving, do I need to bump **tcs-parent-client** too?


hihilary:
Can't find `tcs-parent-client` in this project, but if what you meant is `tis-parent-client`, then we don't need to bump it as it's in shared-modules.

hihilary (approved):

PR #823, opened 2022-01-28
fix: programmeMembership leavingDestination

ReubenRobertsHEE:
NIMDTA is fine, this is only for HEE.

TIS21-2600


Judge40 (approved):

hihilary:
I found these 2 programmeMemberships has already had the `leavingDestination` 'Completed Training'.
image

ReubenRobertsHEE:
@hihilary that's right - I just checked, they must have been fixed manually since I last pulled down a copy of the DB.

PR #824, opened 2022-02-01
chore: fix trainingPathway data of ProgrammeMembership

hihilary:
Now, on TIS:
the trainingPathway column of **92969** programmeMembership records need to be set `CCT` 
the trainingPathway column of **141615** programmeMembership records need to be set `N/A` 

on NIMDTA:
the trainingPathway column of **25** programmeMembership records need to be set `CCT` 

Due to limit of `range_optimizer_max_mem_size`, the super long Id list is split and every update SQL line contains 1000 records.

TIS21-2381


ReubenRobertsHEE:
Do you think its necessary to set values where they are already the desired value? I'm not sure if that is being done to avoid possible inconsistency through user's editing individual values within a programme membership 'cluster' - though I was under the impression this was not possible through the UI. It probably makes the queries slower and bulkier than they might need to be.
Example: the first line of V4.13 updates: `UPDATE `ProgrammeMembership` SET `trainingPathway` = 'N/A' where `id` in (... 142, ...);` but this record already has this value (https://build.tis.nhs.uk/metabase/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7InR5cGUiOiJxdWVyeSIsInF1ZXJ5Ijp7InNvdXJjZS10YWJsZSI6MTcyMywiZmlsdGVyIjpbImFuZCIsWyI9IixbImZpZWxkLWlkIiwyNjIzNl0sMTQyXV19LCJkYXRhYmFzZSI6MzZ9LCJkaXNwbGF5IjoidGFibGUiLCJ2aXN1YWxpemF0aW9uX3NldHRpbmdzIjp7fX0=) 
There are lots of these

ReubenRobertsHEE:
Ah, I think I see the problem @hihilary - is the first set of lines for V4.13 supposed to have 'UPDATE ProgrammeMembershipSETtrainingPathway= 'CCT' (not 'N/A')  ?

hihilary:
> Ah, I think I see the problem @hihilary - is the first set of lines for V4.13 supposed to have 'UPDATE ProgrammeMembershipSETtrainingPathway= 'CCT' (not 'N/A') ?

That's right Reuben, just pushed a commit to fix it. I used some script to generate the SQL scripts but carelessly forgot to update that value.

ReubenRobertsHEE (approved):
Looks good!

PR #826, opened 2022-02-02
chore: fix trainingPathway data of ProgrammeMembership

hihilary:
#824 deployment failed may be because the time limit

~~1. update the Id list as programmeMemberships are changing on prod~~
2. temperarily set `wait-for-minutes` of TIS Stage/Prod to **60** mins to allow the workflow wait for longer

~~If `wait-for-minutes` still doesn't work,~~ I will try setting the `health_check_grace_period` of tcs service.

TIS21-2381


.github/workflows/ci-cd-workflow.yml
```diff
@@ -192,6 +192,7 @@ jobs:
           service: tis-tcs
           cluster: tis-preprod
           wait-for-service-stability: true
+          wait-for-minutes: 60
```
Judge40, line 195: I'm not sure we should change this.
If the start-up takes too long the GHA will fail, but the script will still complete in the running service.

Redeploy would then deploy in a "normal" amount of time.
Bit of a faff having to re-deploy a couple of times to get everything green, but I'm not sure I'd want such a large delay before knowing that a failed deploy really did fail.

.github/workflows/ci-cd-workflow.yml
```diff
@@ -229,6 +230,7 @@ jobs:
           service: tis-tcs
           cluster: tis-prod
           wait-for-service-stability: true
+          wait-for-minutes: 60
```
Judge40, line 233: As above.

.github/workflows/ci-cd-workflow.yml
hihilary: Yesterday, we had 2 attempts of deployments: https://github.com/Health-Education-England/TIS-TCS/runs/5020498400?check_suite_focus=true. Both failed, and then I checked the data, the script seemed not to be completed. 
So I'm wondering if it's okay to increase this time limit **temporarily** and have a try. I will change it back to default after.

Judge40: This property doesn't affect deployment success or failure, it is just a timed [waitFor](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/ECS.html#waitFor-property). After the timer was expired the workflow run will be flagged as failed but it does not revert/cancel the ECS deployment.

hihilary: Okay, I've got rid of this attribute.

Judge40 (approved):

PR #841, opened 2022-03-11
feat: add validation of Placement Specialties

hihilary:
1. There's only 1 primary specialty allowed and it's required. There's only 1 sub specialty allowed.
2. All the specialties should exists in Specitly table.
3. If the sub specialty is specified, it should be from the parent post.

TIS21-2356


hihilary:
Found the specialtyName is missing when UI/bulk upload sends placement to TCS, which should be fixed in UI/bulk upload (I think).

Judge40 (approved):

PR #842, opened 2022-03-16
Feat: event listeners use curriculum membership

ReubenRobertsHEE:
TIS21-2751


ReubenRobertsHEE:
It might be that these changes should be minor version changes, not bug fixes.

Something which might be worth doing now as well would be to rename the ProgrammeMembershipService (which uses the curriculum membership repository) to CurriculumMembershipService, since its a bit confusing as it stands. The reason this has not just been done is that there is a fair number of variables still named ProgrammeMembership that could also be refactored - I guess it should be done, it will just make a PR that is awkward to review...

hihilary:
> It might be that these changes should be minor version changes, not bug fixes.
> 
> Something which might be worth doing now as well would be to rename the ProgrammeMembershipService (which uses the curriculum membership repository) to CurriculumMembershipService, since its a bit confusing as it stands. The reason this has not just been done is that there is a fair number of variables still named ProgrammeMembership that could also be refactored - I guess it should be done, it will just make a PR that is awkward to review...

IMHO renaming ProgrammeMembershipService probably can be left until we rebuild ProgrammeMembership table. Because we probably need to rewrite this service to use the new structure.
When rebuild the ProgrammeMembership, we might need to rewrite the save/update/delete and keep queries from the CurriculumMembership table. Then rewrite queries gradually. 
I guess we might still need both ProgrammeMembership & CurriculumMembership servcie, but should figure out what should go to which.

PR #843, opened 2022-03-17
feat: event listeners use CurriculumMembership

ReubenRobertsHEE:
TIS21-2751


JosephKelly (approved):
Yeah,  I'm ok about the duplication.  I'm not sure how much we'd gain from having an abstract `CurriculumEvent`.  There would be the typical small improvement in having a single `equals()` method to test etc. but it's v. small.

PR #847, opened 2022-03-31
feat: Capture Post ESR events

ReubenRobertsHEE:
API endpoint to receive these from EsrInboundDataWriter, and new database
table PostEsrEvent to record them into.

TIS21-2636: create end point in TCS


tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/Post.java
```diff
@@ -87,6 +87,8 @@
   private Set<Placement> placementHistory = new HashSet<>();
   @OneToMany(mappedBy = "post", fetch = FetchType.LAZY)
   private Set<PostTrust> associatedTrusts;
+  @OneToMany(mappedBy = "post", fetch = FetchType.EAGER, cascade = CascadeType.REMOVE)
```
hihilary, line 90: I'm a little wondering if the FetchType can be set to Lazy because I saw the `placementEsrEvents` in Placement class is.

tcs-client/pom.xml
```diff
@@ -37,7 +37,7 @@
     <dependency>
       <groupId>com.transformuk.hee.tis</groupId>
       <artifactId>tcs-api</artifactId>
-      <version>6.14.4</version>
+      <version>6.15.0</version>
```
hihilary, line 40: Just think if the dependencies are updated, then the module itself might need a version bump too.

tcs-persistence/pom.xml
```diff
@@ -18,7 +18,7 @@
     <dependency>
       <groupId>com.transformuk.hee.tis</groupId>
       <artifactId>tcs-api</artifactId>
-      <version>6.14.4</version>
+      <version>6.15.0</version>
```
hihilary, line 21: tcs-persistence might need a version bump too.

hihilary (commented):
The tcs-service needs a version bump.

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/Post.java
ReubenRobertsHEE: Good point, let me do a bit of checking 🔬

tcs-client/pom.xml
ReubenRobertsHEE: Sure thing :)

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/Post.java
```diff
@@ -86,7 +87,10 @@
   @OneToMany(mappedBy = "post", fetch = FetchType.LAZY)
   private Set<Placement> placementHistory = new HashSet<>();
   @OneToMany(mappedBy = "post", fetch = FetchType.LAZY)
+  @JsonManagedReference
```
ReubenRobertsHEE, line 90: This was not actually related to the fetch type for Posts (LAZY or EAGER) but something I noticed was causing some infinite recursion errors in JSON serialization, e.g.
```
Could not write JSON: Infinite recursion (StackOverflowError); nested exception is com.fasterxml.jackson.databind.JsonMappingException: Infinite recursion (StackOverflowError) (through reference chain: com.transformuk.hee.tis.tcs.service.model.Post["associatedTrusts"]->
org.hibernate.collection.internal.PersistentSet[0]->
com.transformuk.hee.tis.tcs.service.model.PostTrust["post"]->
com.transformuk.hee.tis.tcs.service.model.Post["associatedTrusts"]->
```
This may require review and/or more testing to be added

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/PostTrust.java
```diff
@@ -25,6 +26,7 @@
 
   @ManyToOne(fetch = FetchType.LAZY, optional = false)
   @JoinColumn(name = "postId")
+  @JsonBackReference
```
ReubenRobertsHEE, line 29: See comment on Post object above.

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/Post.java
ReubenRobertsHEE: Ah, this doesn't seem to affect things (though I'd agree LAZY fetch is better, all other things being equal). But I did notice something else in the JSON serialization - see Post.java comment below!

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/Post.java
hihilary: May I ask how I can replicate this issue, please?
There're many 1 to many mapping properties in the Post class, so just a little wondering if this happens on others.

ReubenRobertsHEE: It happens when the ESR endpoint is called, but in debugging it only throws the error on the second or subsequent call (/posts/{postId}/esr-reconciled) - let me try to find a slightly simpler use case to set up, because that's very clumsy!

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
```diff
@@ -629,6 +637,23 @@ public void canLoggedInUserViewOrAmend(Long postId) {
     return postMapper.postsToPostDTOs(result);
   }
 
+  @Override
+  public Optional<PostEsrEvent> markPostAsEsrPositionChanged(Long postId,
+                                                             PostEsrEventDto postEsrExportedDto) {
+    Optional<Post> optionalPostId = postRepository.findPostWithTrustsById(postId);
```
ReubenRobertsHEE, line 643: This seems to be the line that triggers the infinite recursion

hihilary (approved):

PR #851, opened 2022-04-04
data: populate PostEsrEvent table

ReubenRobertsHEE:
Data drawn from ESrReconciliation - Prod as of 2022-04-04T11:21 BST.

Only records with Status=MATCHED included. Status set to DELETED if
Deleted=true, otherwise RECONCILED.

TIS21-2696


JosephKelly:
> The eventDateTime is shown as 'reconciled Date' on Admins-UI, but there're no other timestamps we can get from the ESRReconciliation table. I'm wondering for all the historical data, if showing '04/04/2020' directly to the users is okay.

Good point.  Much like the filename, one suggestion I can make is to use a 'LOW' value, so any message received will be later.  `1000-01-01 00:00:00` seems like the minimum for a MySQL `DATETIME`, which is in range for a Java `LocalDateTime`.

hihilary:
> Thinking ahead to unknown possibles, a `TIMESTAMP`/epoch compatible time might be better: `1970-01-01 00:00:01`.

`1970-01-01 00:00:01` sounds good to me. But I'm thinking if it's better to get a confirmation from James? Because it's also exported to NDW, and I probably need to know what to present to the users :)

hihilary (approved):

PR #852, opened 2022-04-05
fix: refactor PostEsrEvent population

ReubenRobertsHEE:
Much more efficient: local run time drops from ~200sec to <4sec for the
new version.

TIS21-2686: Create records in TIS for positions that are already
reconciled


ReubenRobertsHEE:
The previous deployment failed: I can only assume it timed-out. This version inserts 100 rows at a time (which I should have done up-front, sorry!)

hihilary (approved):

PR #853, opened 2022-04-05
fix: ignore missing Posts when populating table

ReubenRobertsHEE:
The flyway script was built from Prod data, but Staging seems to be
missing some Posts, causing errors. This should convert these to
warnings.

TIS21-2696: Create records in TIS for positions that are already
reconciled


ReubenRobertsHEE:
Sorry, I'm an idiot - I assumed staging was up-to-date with Prod data :/

JosephKelly (approved):

PR #854, opened 2022-04-06
fix: email address validation regexp

ReubenRobertsHEE:
This makes the email address validation regexp consistent with the one
used in admins-ui.

Note that it is not comprehensive: tests have been enhanced to
demonstrate failing examples. As such, the TODO for finding a better
solution has been left in place.

TIS21-2764

### DIFF
--- a/tcs-service/pom.xml
+++ b/tcs-service/pom.xml
@@ -12,7 +12,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-service</artifactId>
-  <version>6.24.1</version>
+  <version>6.24.2</version>
 
   <dependencies>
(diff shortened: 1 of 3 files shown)


ReubenRobertsHEE:
This is a bit of cheeky fix (i.e. it is not very good, but it is 'quick' 😃 ). If we prefer a more comprehensive solution, I think we'd need to review the validation in admins-ui and generic-upload as well to make sure everything ends up consistent. I was not sure of the immediate value of that work, hence the limited PR.

Judge40 (approved):
Just going to approve this as I'm not going to try mentally parse the regex, the tests show that its an improvement and its in keeping with the validation used elsewhere.

PR #860, opened 2022-04-19
Fix: Handle NPEs

EdwardBarclay:
Hotfix for ticket TIS21-505

NPE encountered when users updated either the VisaValidTo date or the VisaIssued date when none present in the DB

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/RightToWorkValidator.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/RightToWorkValidator.java
@@ -8,12 +8,14 @@ import com.transformuk.hee.tis.tcs.api.enumeration.Settled;
 import com.transformuk.hee.tis.tcs.service.model.Person;
 import com.transformuk.hee.tis.tcs.service.model.RightToWork;
 import com.transformuk.hee.tis.tcs.service.repository.PersonRepository;
+
 import java.lang.reflect.Method;
 import java.time.LocalDate;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+
 import org.springframework.core.MethodParameter;
 import org.springframework.stereotype.Component;
 import org.springframework.validation.BeanPropertyBindingResult;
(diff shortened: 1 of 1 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/RightToWorkValidator.java
```diff
@@ -90,35 +90,52 @@ private void checkSettled(RightToWorkDTO dto, List<FieldError> fieldErrors) {
...
-                  + "before current visaIssued date.");
+              new FieldError(DTO_NAME, FIELD_NAME_VISA_ISSUED, "visaIssued must be"
+                  + " before visaValidTo.");
           fieldErrors.add(fieldError);
         }
+      } else {
+        dbValidationError = checkDbVisaDates(personId, visaIssued, visaValidTo);
+      }
+
+      if (dbValidationError) {
```
JosephKelly, line 117: You could condense this to `} else if (checkDbVisaDates(personId, visaIssued, visaValidTo)) {` and get rid of the `dbValidationError` variable.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/RightToWorkValidator.java
```diff
@@ -90,35 +90,52 @@ private void checkSettled(RightToWorkDTO dto, List<FieldError> fieldErrors) {
...
+        FieldError fieldError =
+            new FieldError(DTO_NAME, FIELD_NAME_VISA_ISSUED, "Visa Dates conflict"
+                + " with dates already in Database");
```
JosephKelly, line 120: The error might be with the `visaValidTo` passed in rather than `visaIssued`.

You could inline the method if it doesn't break cognitive complexity constraints.  If it does you could 1)still inline the checksuse DB values to set whichever was null of `visaValidTo` or `visaIssued` and then do the compare, or 2)pass in `fieldErrors` to `checkDbVisaDates()`.

JosephKelly (changes requested):

hihilary:
The `checkVisaDates()` validation has been added for `validateForBulk()` which is for bulk upload only.
We probably need to add it in `validate()` to check the API calls.

JosephKelly (approved):

PR #863, opened 2022-04-21
feat: extend the validation for visa to validate() in RightToWorkvali…

EdwardBarclay:
…dator

Ticket TIS21-505


JosephKelly (approved):
Work checking with a BA/Data why we don't validate everything for this like with bulk upload!

EdwardBarclay:
> Work checking with a BA/Data why we don't validate everything for this like with bulk upload!

I was Talking with Yafang and she said that when the bulk upload functionalities were added that making it consistent was not seen as very urgent, so you will find that validate() is lacking in verifications that are in the bulk upload validation. but if we are adding new ones its best to have it on both.

PR #867, opened 2022-04-25
fix: visaDates validation

hihilary:
fix:
1. unnecessary parameters (personId)
2. a swtich `checkExistingDBValues` to control if we need to check DB values for the update
3. missing validation "visaIssued should be no later than visaValidTo“ in `checkDbVisaDates`
4. add tests

Why do I need to add the swich `checkExistingDBValues`?
When the visa dates are updated from frontend, it will be a replacement and any empty value should be treated as a valid value.
While for bulk upload, the current logic is to ignore the empty values in the template, and only the values which are not empty will be updated. For people bulk create (this can be used as an update too), the generic-upload service retrieves the DB record and only update the non-empty fields and send the assembled entity to tcs. For people bulk update, the process of assembly is after the data validation.

TIS21-505

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/RightToWorkValidator.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/RightToWorkValidator.java
@@ -91,7 +91,8 @@ public class RightToWorkValidator {
     }
   }
 
-  private void checkVisaDates(RightToWorkDTO dto, List<FieldError> fieldErrors, boolean checkExistingDBValues) {
+  private void checkVisaDates(RightToWorkDTO dto, List<FieldError> fieldErrors,
+                              boolean checkExistingDbValues) {
 
     if (dto != null) {
       LocalDate visaIssued = dto.getVisaIssued();
@@ -110,7 +111,7 @@ public class RightToWorkValidator {
                   + " before visaValidTo.");
           fieldErrors.add(fieldError);
         }
-      } else if (checkExistingDBValues){
+      } else if (checkExistingDbValues) {
         checkDbVisaDates(fieldErrors, dto.getId(), visaIssued, visaValidTo);
       }
     }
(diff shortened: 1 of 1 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/RightToWorkValidator.java
```diff
@@ -46,14 +46,14 @@ public class RightToWorkValidator {
    * @param dto the rightToWork to check
    * @throws MethodArgumentNotValidException if there are validation errors
    */
-  public void validate(RightToWorkDTO dto, Long personId) throws MethodArgumentNotValidException {
```
hihilary, line 49: dto.getId() == personId, so it's unnecessary

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/RightToWorkValidator.java
```diff
@@ -110,25 +111,36 @@ private void checkVisaDates(List<FieldError> fieldErrors, RightToWorkDTO dto, Lo
...
 
+    if (personId == null) {
+      return;
+    }
+
     Optional<Person> originalPersonRecord = personRepository.findPersonById(personId);
 
     if (originalPersonRecord.isPresent()) {
       RightToWork oldRtwDto = originalPersonRecord.get().getRightToWork();
-      if (visaIssued != null && oldRtwDto.getVisaValidTo() != null) {
```
hihilary, line 126: No actual date validation here.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/RightToWorkValidator.java
```diff
@@ -110,25 +111,36 @@ private void checkVisaDates(List<FieldError> fieldErrors, RightToWorkDTO dto, Lo
...
+      }
+      LocalDate oldVisaIssued = oldRtwDto.getVisaIssued();
+      LocalDate oldVisaValidTo = oldRtwDto.getVisaValidTo();
+
+      if (visaIssued != null && oldVisaValidTo != null && visaIssued.isAfter(oldVisaValidTo)) {
         FieldError fieldError =
             new FieldError(DTO_NAME, FIELD_NAME_VISA_ISSUED, "Visa Issued Date "
                 + "conflicts with Visa Valid to date already in Database");
         fieldErrors.add(fieldError);
-      } else if (visaValidTo != null && oldRtwDto.getVisaIssued() != null) {
```
hihilary, line 131: No actual date validation here too.

JosephKelly (approved):

PR #874, opened 2022-05-20
fix: get latest curriculum membership based on curriculum end date

jayanta2018:
In order to build the connection dtos  previously it was pulling latest CurriculumMembership based on programmeEndDate but the correct one should be  get latest CurriculumMembership based on curriculumEndDate


DorisWongNHS:
The code LGTM. Do we need to update the version?

jayanta2018:
> 

Good shout Doris. I totally forgot about that. :laughing:

ReubenRobertsHEE:
Do you think it would be worth including a test that shows the new code picks out the right record, if a pair of curriculumMembership records are available, where the ordering of the pair of records by programmeEndDate is different from the ordering by curriculumEndDate?

hihilary:
I'm not very familiar with the requirement. This PR looks that `buildTcsConnectionInfo` method uses the ProgrammeMembership fields (including programme startDate & endDate) of the latest CurriculumMembership (ordered by curriculum endDate). But the curriculumEndDate is not in the returned `ConnectionInfoDto`.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/RevalidationServiceImpl.java
```diff
@@ -249,7 +249,9 @@ public ConnectionInfoDto buildTcsConnectionInfo(Long personId) {
           .programmeName(programmeMembershipDto.getProgrammeName())
           .programmeMembershipStartDate(programmeMembershipDto.getProgrammeStartDate())
           .programmeMembershipEndDate(programmeMembershipDto.getProgrammeEndDate())
+          .curriculumEndDate(latestCurriculumMembership.getCurriculumEndDate())
           .programmeMembershipType(membershipType != null ? membershipType.toString() : null);
+
```
hihilary, line 254: This blank line needs to be removed.

jayanta2018: ok done @hihilary

hihilary (approved):

PR #875, opened 2022-05-20
fix: NIMDTA-aware Programme Membership refactor

ReubenRobertsHEE:
There are some records with NULL programmeStartDates which broke the
previous flyway V4.21_refactor_programmeMembership.sql script

The new checksum is -541714420

TIS21-2510: Rebuild ProgrammeMembership table and populate


JosephKelly (approved):
LGTM :smile:

PR #876, opened 2022-06-08
fix: trim and remove all non-breaking spaces in emails

hihilary:
1. emails shouldn't have any non-breaking spaces, so replace all of them with ''
2. the setting 'sql_safe_updates' are off on Stage/Prod/NIMDTA, so it's okay not to have ID in where clause for updates.
3. records affected: Prod: 108, NIMDTA: 4.

TIS21-2866


Judge40 (commented):
🥳

PR #877, opened 2022-06-09
test: add tests for email validation

hihilary:
Before we applied the flyway script to fix the email on Prod/NIMDTA, the latest email needed to be fixed had the amendedDate "2021-10-12T14:42:23.155", and we fixed the RegEx of email on 06/04/2022. By adding the tests, I found the whitespaces and non-breaking spaces are already filtered out by the new RegEx.

Besides, according to [Wikipedia](https://en.wikipedia.org/wiki/Non-breaking_space), the non-breaking space in Unicode is `U+00A0`.

TIS21-2866

### DIFF
--- a/tcs-service/pom.xml
+++ b/tcs-service/pom.xml
@@ -12,7 +12,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-service</artifactId>
-  <version>6.26.0</version>
+  <version>6.26.1</version>
 
   <dependencies>
     <dependency>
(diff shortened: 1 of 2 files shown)


JosephKelly (approved):

hihilary:
> Ah, but the ContactDetails table is utf8 encoded, not unicode afaik ;)

Yeah, the character coding in our MySQL is utf-8, but the native character encoding of the Java programming language is UTF-16. The UTF-8(hex) C2A0 and UTF-16 00A0 are the same in the memory, Java doesn't care what the character coding of the database. When the string is read, Java converts it to its own coding defined by JVM.

I've also tested the pre-fix data, the non-breaking space equals u+00A0 as expected.

PR #888, opened 2022-08-25
fix: get personStatus updated when a programmeMembership is deleted

hihilary:
Cause of the bug:
The entities Person and ProgrammeMembership are 1 to many mapping (bidirectional relationship), and when a programmeMembership is deleted from the child side, the deletion doesn't get synchronised to the parent side. 
Some similar discussion can be found here: https://stackoverflow.com/questions/22688402/delete-not-working-with-jparepository

Solutions I can think of and have verified on local:
1. get the changes flushed to the DB before populating personStatus, so that (I think) the cached person object will get updated.
2. add a `@PreRemove` hook to dimiss the parent before the child programmeMembership is deleted.

I've chosen the 2nd solution as it doesn't break the transaction itself and solves the real problem.

TIS21-3311


JosephKelly (approved):
Code looks good to me.  I'm happy enough that you've looked into it & don't understand the tx boundaries.

I've not addressed the other smells in this PR to prevent other classes being further inspected.

PR #895, opened 2022-10-13
Feat/fetch current pm for reval sync

hihilary:
1. only export current PM with the latest CM
2. if there're multiple current PMs, the `programmeName` is set to "multiple programmes" and `programmeStartDate`, `programmeEndDate`, `programmeMembershipType` are empty
3. if there're no current PM, `programmeStartDate`, `programmeEndDate`, `programmeMembershipType` are empty

The total line numbers of this query is one less than the count of Persons, as 1 person (id: 5893) doesn't have ContactDetails.
In my local (debug mode), the excute time of `extractConnectionInfoForSync()` is around 25s.

### DIFF
--- a/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
+++ b/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
@@ -5,25 +5,34 @@ from (
     gmc.gmcNumber,
     cd.forenames,
     cd.surname,
-    latestPm.programmeMembershipType,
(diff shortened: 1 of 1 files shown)


JosephKelly (approved):
Interesting approach.  Good enough for me.  If you looked at the efficiency of this vs. a `case select` and a deeper tree of queries.  I'm thinking ahead to where we might want to use this query for a single `personId`.

I think this is more readable.

PR #896, opened 2022-10-17
feat: add clauses For Reval Sync Sql

hihilary:
I did "EXPLAIN" traineeInfoForConnection.sql, I think the performance are decided by the rows of Person, nested query to get current PMs and nested query to get max CM.

So I tried limiting the query rows of personId to 10000, and the below pics are comparison of EXPLAIN output when I added _where clauses_ in different places:

- Add where clause in `ot` only:

image
https://build.tis.nhs.uk/metabase/question/474

- Add where clauses in nested queries and Person table:

image
https://build.tis.nhs.uk/metabase/question/473

We can see the 1st query needs to examine more rows to execute the whole query.

Some docs of "EXPLAIN": https://dev.mysql.com/doc/refman/5.7/en/explain-output.html

TIS21-3229

### DIFF
--- a/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
+++ b/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
@@ -21,6 +21,7 @@ from (
       personId,
       count(if(pm.programmeStartDate <= current_date() and pm.programmeEndDate >= current_date(), true, null)) as count_num
     from ProgrammeMembership pm
+    WHERECLAUSE1 -- where personId in ()
     group by personId
   ) currentPmCounts on p.id = currentPmCounts.personId
   left join ProgrammeMembership pm1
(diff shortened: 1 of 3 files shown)


tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
```diff
@@ -31,8 +32,12 @@ from (
           cm.programmeMembershipUuid,
           max(cm.curriculumEndDate) curriculumEndDate
       from CurriculumMembership cm
+      WHERECLAUSE1 -- where personId in ()
       group by cm.programmeMembershipUuid
   ) latestCm on latestCm.programmeMembershipUuid = pm1.uuid
   left join Programme prg on prg.id = pm1.programmeId
+  WHERECLAUSE2 -- where p.id in ()
```
JosephKelly, line 39: Looks good & comments are really useful.

I'd check this before committing but: I think you could replace `WHERECLAUSE2` by modifying traineeInfoForConnection.sql#L4 with `   p.id personId`.  There may be something to change in the Java to match the alias but if it improves the readability without hurting performance then it's a positive difference.

JosephKelly (approved):
LGTM

PR #899, opened 2022-10-18
feat:use RevalSyncSql for Single Person export

hihilary:
1. [The commit](https://github.com/Health-Education-England/TIS-TCS/pull/896/commits/a4dd1e1ad2f71f424bada7f2f0184100be266467) which uses alias `personId` for the 3rd where clause doesn't work. (I thought I had tested it very well, but apparently not😢)Because: 

- personId exists in temporary tables `latestCm` and `currentPmCounts`, so it's ambiguous for where clause; 
- MySQL executes where clauses first, then maps aliases, so personId is not available for p.id when where clause is being executed, unless we use `having`. But using `having` means that we will lose the optimisation of the DB and results in bad performance.

2. add macros in the format of `WHERECLAUSE(p, id)` in the SQL, so that we can recognise the macro by RegEx and get the parameters from the macro and populate the real where clauses.

3. apply the SQL to export a single ConnectionInfoDto.

TIS21-3229

### DIFF
--- a/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
+++ b/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
@@ -21,7 +21,7 @@ from (
       personId,
       count(if(pm.programmeStartDate <= current_date() and pm.programmeEndDate >= current_date(), true, null)) as count_num
     from ProgrammeMembership pm
-    WHERECLAUSE
+    WHERECLAUSE(pm, personId)
     group by personId
   ) currentPmCounts on p.id = currentPmCounts.personId
   left join ProgrammeMembership pm1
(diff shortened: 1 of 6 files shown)


JosephKelly:
Looking at the base query.  I don't see what it is that the Person table is doing that another table couldn't do.
I'm not certain yet which table would be the primary search table but we could just get the `personId` from that table.

hihilary:
> Looking at the base query. I don't see what it is that the Person table is doing that another table couldn't do. I'm not certain yet which table would be the primary search table but we could just get the `personId` from that table.

@JosephKelly good spot. 
Below are my thinking:
1. `Person` table should have the same count as `ContactDetails`, but there's 1 record which is in `Person` doesn't exist in `ContactDetails`: https://build.tis.nhs.uk/metabase/question/475. I've no idea how this happens, but in the original SQL we use `join`, so this incomplete record is already ignored.
2. Not every person has Gmc ID.
3. Not every person has PMs or CMs. 

So my conclusion is that we can get rid of `Person` table and use `ContactDetails` as primary search table.

jayanta2018 (commented):
As we moved from service layer to DB, I wonder there might be some need to write integration tests by involving repository layer. @hihilary

JosephKelly (approved):
Got that extra reassurance from @CaiWillis.  Thanks!

This has an effect as soon as it's deployed so might not be ready to be merged.

hihilary:
> Got that extra reassurance from @CaiWillis. Thanks!
> 
> This has an effect as soon as it's deployed so might not be ready to be merged.

@JosephKelly  This PR should fix the data for ongoing changes of PM. 
How about merging it and testing on Stage, and stop it to be pushed to prod for now?

PR #902, opened 2022-10-26
feat:modify sql to use combined current programmes

hihilary:
for Reval sync

TIS21-3229

### DIFF
--- a/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
+++ b/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
@@ -1,7 +1,7 @@
 select distinct ot.*
 from (
   select
-    cd.id personId,
+    cd.id as personId,
     gmc.gmcNumber,
     cd.forenames,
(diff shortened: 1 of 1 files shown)


tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
```diff
@@ -1,40 +1,43 @@
...
     ContactDetails cd
   left join GmcDetails gmc on (gmc.id = cd.id)
   left join (
-    -- count current PMs for each person
+    -- count current PMs with combined programme names for each person
     select
-      personId,
+      pm.personId,
+      GROUP_CONCAT(prg.programmeName SEPARATOR " | ") programmeNames,
+      GROUP_CONCAT(prg.owner SEPARATOR " | ") owner,
```
hihilary, line 22: If we don't export `owner` here, we will need to left join `Programme` again with the primary search.

jayanta2018 (approved):

PR #914, opened 2023-01-16
refactor: trainingNumberEsEventListener to use ProgrammeMembershipRepository.

hihilary:
The initial refacotring from PM to CM is in this commit: https://github.com/Health-Education-England/TIS-TCS/commit/3fd3540

`TrainingNumber` is deprecated in CM, so we have to change the listener to use PM.

However, there's no entry to modify training numbers on TIS admin UI. And, in tcs-service, I'm not able to find anywhere who sends the `TrainingNumberXXXEvent`. Had a quick discussion with James before, and he also mentioned this functionality would be added in the future.

In this PR, I added a query in `ProgrammeMembershipRepository` as I thought we need it, but the current process doesn't modify the `trainingNumber` in PMs, which needs to be fixed when we start to implement this functionality.

TIS21-2388

### DIFF
--- a/tcs-persistence/src/test/java/com/transformuk/hee/tis/tcs/service/repository/ProgrammeMembershipRepositoryIntTest.java
+++ b/tcs-persistence/src/test/java/com/transformuk/hee/tis/tcs/service/repository/ProgrammeMembershipRepositoryIntTest.java
@@ -70,7 +70,7 @@ public class ProgrammeMembershipRepositoryIntTest {
   @Test
   public void shouldFindByTrainingNumberId() {
     List<ProgrammeMembership> programmeMemberships =
-        programmeMembershipRepository.findByTrainingNumber_Id(trainingNumber1.getId());
+        programmeMembershipRepository.findByTrainingNumberId(trainingNumber1.getId());
     Assert.assertEquals(2, programmeMemberships.size());
   }
 }
--- a/tcs-persistence/src/test/java/com/transformuk/hee/tis/tcs/service/repository/ProgrammeMembershipRepositoryIntTest.java
+++ b/tcs-persistence/src/test/java/com/transformuk/hee/tis/tcs/service/repository/ProgrammeMembershipRepositoryIntTest.java
@@ -1,0 +1,76 @@
+package com.transformuk.hee.tis.tcs.service.repository;
+
+import com.transformuk.hee.tis.tcs.service.TestConfig;
+import com.transformuk.hee.tis.tcs.service.model.Person;
+import com.transformuk.hee.tis.tcs.service.model.Programme;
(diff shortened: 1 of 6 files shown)


tcs-persistence/src/test/java/com/transformuk/hee/tis/tcs/service/repository/ProgrammeMembershipRepositoryIntTest.java
```diff
@@ -0,0 +1,76 @@
+package com.transformuk.hee.tis.tcs.service.repository;
```
jayanta2018, line 1: It is a new file. Just wondering do we need to have the license declaration at the beginning?

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/listener/person/TrainingNumberElasticSearchEventListenerTest.java
```diff
@@ -0,0 +1,108 @@
+package com.transformuk.hee.tis.tcs.service.listener.person;
```
jayanta2018, line 1: May be the same here

tcs-persistence/src/test/java/com/transformuk/hee/tis/tcs/service/repository/ProgrammeMembershipRepositoryIntTest.java
hihilary: I can't see any license declaration in any of other tcs files, so think it's not needed.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/listener/person/TrainingNumberElasticSearchEventListenerTest.java
hihilary: I can't see any license declaration in any of other tcs files, so think it's not needed.

hihilary:
> I have looked for the rationale for changing from searching `curriculumMembershipRepository.findByTraineeId()` to `programmeMembershipRepository.trainingNumberId()` rather than `programmeMembershipRepository.findByPersonId()` but couldn't see it.
> 
> Based on the data model it looks like a bug but I'm not certain.

TrainingNumber is a separate table in tcs (like some metadata), and every ProgrammeMembership has a nested TrainingNumber object. (CurriculumMembership.TrainingNumber is deprecated)
I think the initial purpose of this listener should be when a training number details is modified, all the nested TrainingNumbers in ProgrammeMembership have to be modified accordingly.

So what I've done here is to get the ProgrammeMemberships which needs a modification. However, currently, we don't modify anything.

JosephKelly (approved):
Cool.  Squashing probably bugs too :-)

PR #917, opened 2023-02-22
refactor: rename attribute names

jayanta2018:
TIS21-3774

### DIFF
--- a/tcs-persistence/pom.xml
+++ b/tcs-persistence/pom.xml
@@ -18,7 +18,7 @@
     <dependency>
       <groupId>com.transformuk.hee.tis</groupId>
       <artifactId>tcs-api</artifactId>
-      <version>6.16.1</version>
+      <version>6.16.2</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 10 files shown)


tcs-persistence/pom.xml
```diff
@@ -18,7 +18,7 @@
     <dependency>
       <groupId>com.transformuk.hee.tis</groupId>
       <artifactId>tcs-api</artifactId>
-      <version>6.16.1</version>
+      <version>6.16.2</version>
```
hihilary, line 21: I have a question about the version bump.
The dependency `tcs-api` is updated here, should we bump the version of `tcs-persistence`, and its reference in `tcs-service` to let `tcs-service` know there's a new version?

hihilary (commented):
Another issue I have to raise is, if we decide not to do a full sync from tcs -> Reval, then I assume in ES we will probably have some records which have both `connectionStatus` and `tisConnectionStatus`, some records only have `connectionStatus`.

PR #919, opened 2023-03-06
chore: delete duplicated person record

JosephKelly:
Data recorded against tis id: 265154

TIS21-4266: Merge Trainee Record

### DIFF
--- a/tcs-service/src/main/resources/db/migration/data/hee/V4.24__delete_duplicatedPersonRecord.sql
+++ b/tcs-service/src/main/resources/db/migration/data/hee/V4.24__delete_duplicatedPersonRecord.sql
@@ -1,0 +1,11 @@
+delete from RightToWork where id in(305827);
+-- n/a delete from Qualification where personId in(305827);
+-- n/a delete from ProgrammeMembership where personId in(305827);
+-- n/a delete from PersonTrust where personId in(305827);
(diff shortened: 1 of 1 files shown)


tcs-service/src/main/resources/db/migration/data/hee/V4.24__delete_duplicatedPersonRecord.sql
```diff
@@ -0,0 +1,11 @@
+delete from RightToWork where id in(305827);
```
jayanta2018, line 1: Just wondering is the person id is `305827` or 265154? I think in the app you shared the page with the person id 265154.

JosephKelly: Yep so the one which will remain is `265154` which has more 'actual' data recorded against it

hihilary:
Just had a quick check on the UI. There're more meaningful data under Sensitive data/DIVERSITY. Do we want to copy them over?

JosephKelly:
> Just had a quick check on the UI. There're more meaningful data under "Sensitive data/DIVERSITY" section. Do we want to copy them over?

Thanks for the further check... I'll let the reporter know on Teams & ask them to make sure.

PR #925, opened 2023-03-13
feat: use alternative designated body mapper

JosephKelly:
2 DBCs have been modified in stage but not production.

TIS21-4289: Investigate GMC DBC changes


JosephKelly:
I'm planning on coming back round to this implementation in a couple of days to either:
a) come up with a more elegant solution, or
b) strip it out because we're done testing

hihilary (approved):

PR #926, opened 2023-03-15
chore: remove references to unused field connectionStatus for reval CDC

CaiWillis:
TIS21-4194:TCS, Integration Service & Connection Service - Remove field connectionStatus


JosephKelly (approved):
I'm happy enough with this being a patch but I'm thinking if we were relying on versioning this would be a major change.

hihilary (approved):

PR #929, opened 2023-04-28
feat: store signed COJ against PM ID

Judge40:
Listener for COJ signed events and persist the `signedAt` timestamp and COJ version against the associated ProgrammeMembership UUID.

Ideally this would be persisted via the ProgrammeMembership itself, however there were issues saving the ProgrammeMembershipDto retrieved from the database.
This approach allows us to capture the signed COJ to avoid losing data, given time to come back and resolve the PM persistence when we add the COJ data to the ProgrammeMembershipDto.

TIS21-4431
TIS21-4375

### DIFF
--- a/tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/ConditionsOfJoining.java
+++ b/tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/ConditionsOfJoining.java
@@ -1,0 +1,26 @@
+package com.transformuk.hee.tis.tcs.service.model;
+
+import com.transformuk.hee.tis.tcs.api.enumeration.GoldGuideVersion;
+import java.time.Instant;
+import java.util.UUID;
+import javax.persistence.Entity;
+import javax.persistence.EnumType;
+import javax.persistence.Enumerated;
+import javax.persistence.Id;
+import lombok.Data;
+import org.hibernate.annotations.Type;
+
+/**
+ * A persistable representation of a Conditions of Joining.
+ */
+@Data
+@Entity
+public class ConditionsOfJoining {
+
+  @Id
+  @Type(type = "uuid-char")
+  private UUID programmeMembershipUuid;
+  private Instant signedAt;
+  @Enumerated(EnumType.STRING)
+  private GoldGuideVersion version;
+}
--- a/tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/ConditionsOfJoining.java
+++ b/tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/ConditionsOfJoining.java
@@ -18,7 +18,7 @@ import org.hibernate.annotations.Type;
 public class ConditionsOfJoining {
 
   @Id
-  @Type(type = "uuid-char")
+  @Type(type = "org.hibernate.type.UUIDCharType")
   private UUID programmeMembershipUuid;
   private Instant signedAt;
   @Enumerated(EnumType.STRING)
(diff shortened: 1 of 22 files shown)


tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/ConditionsOfJoining.java
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * A persistable representation of a Conditions of Joining.
+ */
+@Data
+@Entity
+public class ConditionsOfJoining {
+
+  @Id
+  @Type(type = "uuid-char")
```
JosephKelly, line 21: This might be a bit easier on the eyes but is there a reason to have this be different from https://github.com/Health-Education-England/TIS-TCS/blob/ee142daef286ddc2945c5afb7482d8ed1b7090fa/tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/ProgrammeMembership.java#L41?

tcs-service/src/test/resources/db/migration/schema/V4.25__create_conditions_of_joining_table.sql
```diff
@@ -0,0 +1 @@
+/* The uuid field is not available on ProgrammeMembership, so we cannot test this migration. */
```
JosephKelly, line 1: I think I might need you to explain this one to me.  `ProgrammeMembership.uuid` exists in the schema but I don't know why it is not available.

JosephKelly (commented):
Oh man, it took me a little while to wrap my head around why the programme memebership id is sent outside of the `ConditionsOfJoiningDto`.

I'll add this to the 'final' bit of the ProgrammeMembership refactoring work.

I'd like to check the hibernate `@Type` annotation & reason the schema is skipped in tests before approving.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/message/TraineeMessageListener.java
```diff
@@ -0,0 +1,25 @@
...
+@Component
+public class TraineeMessageListener {
+
+  private final ConditionsOfJoiningService conditionsOfJoiningService;
+
+  TraineeMessageListener(ConditionsOfJoiningService conditionsOfJoiningService) {
+    this.conditionsOfJoiningService = conditionsOfJoiningService;
+  }
+
+  @RabbitListener(queues = "${app.rabbit.trainee.queue.coj.signed}", ackMode = "NONE")
```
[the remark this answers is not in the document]
Judge40: I read the docs wrong, it should be `AUTO`

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/ConditionsOfJoining.java
Judge40: None at all, bit of a "this works, `ctrl+c` `ctrl+v`". I'll change it for consistency.

tcs-service/src/test/resources/db/migration/schema/V4.25__create_conditions_of_joining_table.sql
Judge40: When the integration tests run this migration fails as it can't find the `ProgrammeMembership.uuid` in the test database (H2), I had assumed it was due to `v4.21` being commented out too
https://github.com/Health-Education-England/TIS-TCS/blob/ee142daef286ddc2945c5afb7482d8ed1b7090fa/tcs-service/src/test/resources/db/migration/schema/V4.21__refactor_programmeMembership.sql

JosephKelly (approved):

PR #935, opened 2023-05-18
fix: do not requeue COJ sign event when invalid

Judge40:
If a programme is not found for the signed Conditions of Joining an `IllegalArgumentException` is throw. However, this causes the message to be requeued instead of going to the DLQ.
To ensure that the message correctly goes to the DLQ the exception should be caught and rethrown as a `AmqpRejectAndDontRequeueException`.

TIS21-4551
TIS21-4550


DorisWongNHS (approved):
LGTM 👍

JosephKelly (approved):

PR #940, opened 2023-05-30
refactor:mappers of programme/curriculum membersips

hihilary:
refactor some small bits of programme membership and curriculum membership mappers

TIS21-2403

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/mapper/CurriculumMembershipMapper.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/mapper/CurriculumMembershipMapper.java
@@ -18,6 +18,7 @@ import com.transformuk.hee.tis.tcs.service.repository.ConditionsOfJoiningReposit
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.UUID;
 import java.util.stream.Collectors;
 import org.springframework.stereotype.Component;
 import org.springframework.util.CollectionUtils;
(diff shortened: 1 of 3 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/mapper/ProgrammeMembershipMapper.java
```diff
@@ -86,23 +87,23 @@ public List<ProgrammeMembershipDTO> allEntityToDto(
 
   public List<ProgrammeMembershipDTO> programmeMembershipsToProgrammeMembershipDTOs(
       List<ProgrammeMembership> programmeMemberships) {
-    Map<ProgrammeMembershipDTO, ProgrammeMembershipDTO> listMap = Maps.newHashMap();
+    Map<UUID, ProgrammeMembershipDTO> listMap = Maps.newHashMap();
```
hihilary, line 90: Before splitting programme memberships from curriculum memberships, we didn't have a ID, so we had to use the hashcode method in ProgrammeMembershipDTO for hashmap keys.
Now, it would be better to use UUID instead.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/mapper/CurriculumMembershipMapper.java
```diff
@@ -95,23 +90,24 @@ public List<ProgrammeMembershipDTO> allEntityToDto(
    */
   public List<ProgrammeMembershipDTO> curriculumMembershipsToProgrammeMembershipDtos(
       List<CurriculumMembership> curriculumMemberships) {
-    Map<ProgrammeMembershipDTO, ProgrammeMembershipDTO> listMap = Maps.newHashMap();
+    Map<UUID, ProgrammeMembershipDTO> listMap = Maps.newHashMap();
```
hihilary, line 93: Before splitting programme memberships from curriculum memberships, we didn't have a ID, so we had to use the hashcode method in ProgrammeMembershipDTO for hashmap keys.
Now, it would be better to use the UUID instead.

PR #942, opened 2023-06-09
refactor:programme membership

hihilary:
This refactoring is mainly about ProgrammeMembershipMapper:
1. use outside mappers instead of private mapers when possible
2. add more comments, and shorten some method variables to make them more readable
3. in private method `programmeMembershipToProgrammeMembershipDTO`, define an empty curriculumMemberhsip list instead of having to define it in outer layers.

TIS21-2403

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ProgrammeMembershipServiceImpl.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ProgrammeMembershipServiceImpl.java
@@ -61,14 +61,14 @@ public class ProgrammeMembershipServiceImpl implements ProgrammeMembershipServic
   /**
    * Initialise the ProgrammeMembershipService.
    *
-   * @param programmeMembershipRepository   the ProgrammeMembershipRepository
-   * @param curriculumMembershipRepository  the CurriculumMembershipRepository
-   * @param programmeMembershipMapper       the ProgrammeMembershipMapper
-   * @param curriculumMembershipMapper      the CurriculumMembershipMapper
-   * @param curriculumRepository            the CurriculumRepository
-   * @param curriculumMapper                the CurriculumMapper
-   * @param applicationEventPublisher       the ApplicationEventPublisher
-   * @param personRepository                the PersonRepository
+   * @param programmeMembershipRepository  the ProgrammeMembershipRepository
+   * @param curriculumMembershipRepository the CurriculumMembershipRepository
+   * @param programmeMembershipMapper      the ProgrammeMembershipMapper
+   * @param curriculumMembershipMapper     the CurriculumMembershipMapper
+   * @param curriculumRepository           the CurriculumRepository
+   * @param curriculumMapper               the CurriculumMapper
+   * @param applicationEventPublisher      the ApplicationEventPublisher
+   * @param personRepository               the PersonRepository
    */
   public ProgrammeMembershipServiceImpl(
       ProgrammeMembershipRepository programmeMembershipRepository,
@@ -106,12 +106,12 @@ public class ProgrammeMembershipServiceImpl implements ProgrammeMembershipServic
     ProgrammeMembershipDTO programmeMembershipSavedDto
         = programmeMembershipMapper.toDto(programmeMembership);
 
-    //emit events
+    // update Person status
     updatePersonWhenStatusIsStale(programmeMembershipSavedDto.getPerson().getId());
-    List<ProgrammeMembershipDTO> resultDtos
-        = Collections.singletonList(programmeMembershipSavedDto);
-    emitProgrammeMembershipSavedEvents(resultDtos);
-    return CollectionUtils.isNotEmpty(resultDtos) ? resultDtos.get(0) : null;
+    //emit events
+    emitProgrammeMembershipSavedEvents(Collections.singletonList(programmeMembershipSavedDto));
+
+    return programmeMembershipSavedDto;
   }
 
   /**
@@ -132,10 +132,13 @@ public class ProgrammeMembershipServiceImpl implements ProgrammeMembershipServic
       updatePersonWhenStatusIsStale(programmeMembership.getPerson().getId());
     });
 
-    //emit events
     List<ProgrammeMembershipDTO> resultDtos =
         programmeMembershipMapper
             .programmeMembershipsToProgrammeMembershipDTOs(programmeMemberships);
+
+    // update Person status
+    resultDtos.forEach(dto -> updatePersonWhenStatusIsStale(dto.getPerson().getId()));
+    //emit events
     emitProgrammeMembershipSavedEvents(resultDtos);
     return resultDtos;
   }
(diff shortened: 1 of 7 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ProgrammeMembershipServiceImpl.java
```diff
@@ -132,10 +132,13 @@ public List<ProgrammeMembershipDTO> save(List<ProgrammeMembershipDTO> programmeM
       updatePersonWhenStatusIsStale(programmeMembership.getPerson().getId());
     });
 
-    //emit events
     List<ProgrammeMembershipDTO> resultDtos =
         programmeMembershipMapper
             .programmeMembershipsToProgrammeMembershipDTOs(programmeMemberships);
+
+    // update Person status
+    resultDtos.forEach(dto -> updatePersonWhenStatusIsStale(dto.getPerson().getId()));
```
hihilary, line 140: I'm not sure why we don't update person status when doing a bulk saving/updating.

JosephKelly: L#132 looks like we update the person status as soon as we save it.

JosephKelly: We can then refactor this block to nicely remove the `resultDtos` list definition as you did above.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ProgrammeMembershipServiceImpl.java
```diff
@@ -180,8 +183,8 @@ public ProgrammeMembershipDTO findOne(Long id) {
   }
 
   /**
-   * Delete the curriculumMembership by id.
-   * NOTE: this is the curriculumMembership, not the containing programmeMembership
+   * Delete the curriculumMembership by id. NOTE: this is the curriculumMembership, not the
```
JosephKelly, line 186: There are a few things in this method that I'd like to pick up to make it less likely to be misused, including standardising a delete for programme memberships.
Can we talk next week about whether a second refactoring (following this PR) would be useful?

tcs-service/src/main/resources/config/application-local.yml
```diff
@@ -21,6 +21,8 @@ spring:
 
 logging:
     file: ${LOG_DIR:${HOME}}/tcs.log
+    level:
+      root: debug
```
JosephKelly, line 25: Even running locally, logging "debug" by default causes a lot of noise and can make it harder to see application specific logs.
Did you want to set this as the default for TCS?

JosephKelly (changes requested):

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ProgrammeMembershipServiceImpl.java
hihilary: > L#132 looks like we update the person status as soon as we save it.

Ah, I must have got a fuzzy brain....You're right Pepe, will change it back.

tcs-service/src/main/resources/config/application-local.yml
hihilary: Just updated it for some debug level logging, I'm okay to change it back.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/mapper/ProgrammeMembershipMapper.java
```diff
@@ -34,166 +32,187 @@ public class ProgrammeMembershipMapper {
...
-    } else {
-      result.setPerson(personToPersonDTO(programmeMembership.getPerson()));
-    }
-    result.setTrainingPathway(programmeMembership.getTrainingPathway());
-    if (result.getUuid() != null) {
+    dto.setTrainingNumber(
+        trainingNumberMapper.trainingNumberToTrainingNumberDTO(entity.getTrainingNumber()));
+    dto.setPerson(personToPersonDTO(entity.getPerson()));
+
+    dto.setCurriculumMemberships(Lists.newArrayList());
```
JosephKelly, line 207: I noticed that the [docs](https://guava.dev/releases/31.1-jre/api/docs/com/google/common/collect/Lists.html#newArrayList()) say to use the ArrayList constructor in Java 7+

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/mapper/ProgrammeMembershipMapper.java
```diff
@@ -203,17 +222,8 @@ private List<CurriculumMembershipDTO> programmeMembershipToCurriculumMembershipD
...
-      cmDto.setCurriculumStartDate(cm.getCurriculumStartDate());
-      cmDto.setCurriculumEndDate(cm.getCurriculumEndDate());
-      cmDto.setPeriodOfGrace(cm.getPeriodOfGrace());
-      cmDto.setCurriculumCompletionDate(cm.getCurriculumCompletionDate());
-      cmDto.setCurriculumId(cm.getCurriculumId());
-      cmDto.setAmendedDate(cm.getAmendedDate());
-
+      CurriculumMembershipDTO cmDto =
+          curriculumMembershipMapper.curriculumMembershipToCurriculumMembershipDto(cm);
       curriculumMembershipDtos.add(cmDto);
```
JosephKelly, line 227: It looks like this could be simplified even further using the syntax you've used elsewhere:
```java
.stream()
.map(curriculumMembershipMapper::curriculumMembershipToCurriculumMembershipDto)
.collect()
```

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/mapper/ProgrammeMembershipMapper.java
```diff
@@ -34,166 +32,187 @@ public class ProgrammeMembershipMapper {
...
-    for (ProgrammeMembershipDTO programmeMembershipDto : programmeMembershipDTOs) {
-      result.add(toEntity(programmeMembershipDto));
-    }
-
-    return result;
+      List<ProgrammeMembershipDTO> pmDtos) {
+    return pmDtos.stream().map(this::toEntity).collect(Collectors.toList());
   }
 
+  // convert programmeMembership entity to DTO with an empty curriculumMembership list
```
JosephKelly, line 176: It would be good to make clear if/why the curriculum memberships are not added to the programme membership here.  I would have assumed it was a bit of legacy that we didn't want to continue but don't know.

Let me know if I should look into why we'd want to convert the programme membership without the nested curriculum memberships.

JosephKelly (dismissed):

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/mapper/ProgrammeMembershipMapper.java
JosephKelly: I thought methods should be annotated with full "JavaDoc" comments too!

CaiWillis (approved):
Nice and tidy, don't have an answer about `programmeMembershipToProgrammeMembershipDTO`, don't know if you need that sorted before merging

PR #945, opened 2023-06-12
feat: add sentry configuration

jayanta2018:
Sentry configuration has been added.

TIS21-4560

### DIFF
--- a/tcs-service/pom.xml
+++ b/tcs-service/pom.xml
@@ -373,7 +373,7 @@
     <undertow.version>2.0.15.Final</undertow.version>
     <validation-api.version>1.1.0.Final</validation-api.version>
 
-    <sentryVersion>5.3.0</sentryVersion>
+    <sentryVersion>6.22.0</sentryVersion>
   </properties>
 
   <build>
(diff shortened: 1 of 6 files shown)


tcs-service/pom.xml
```diff
@@ -363,6 +372,8 @@
     <tis.shared.modules.version>2.3.0</tis.shared.modules.version>
     <undertow.version>2.0.15.Final</undertow.version>
     <validation-api.version>1.1.0.Final</validation-api.version>
+
+    <sentryVersion>5.3.0</sentryVersion>
```
JosephKelly, line 376: I haven't checked compatibility. Is there any reason the latest version isn't being used?

JosephKelly (approved):
Looks ok, including the use of "stage" as an environment.

I don't think those parameters exist though so should be marked as draft `DO NOT MERGE` until they've been created.

jayanta2018:
> Looks ok, including the use of "stage" as an environment.
> 
> I don't think those parameters exist though so should be marked as draft `DO NOT MERGE` until they've been created.

We will set the environment from Sentry cloud
image

PR #946, opened 2023-06-15
feat: accept CoJ signed events with id or uuid

ReubenRobertsHEE:
For a limited time, events may include a curriculum membership id (long) as previously, or the new programme membership uuid. The latter will be the only accepted form after the transition period.

TIS21-4660

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ConditionsOfJoiningServiceImpl.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ConditionsOfJoiningServiceImpl.java
@@ -44,27 +44,12 @@ public class ConditionsOfJoiningServiceImpl implements ConditionsOfJoiningServic
 
   @Override
   public ConditionsOfJoiningDto save(Object id, ConditionsOfJoiningDto dto) {
-    UUID realProgrammeMembershipId = null;
-    Long curriculumMembershipId = null;
     ProgrammeMembershipDTO programmeMembership;
+    LOG.info("Request received to save Conditions of Joining for id {}.", id);
     try {
-      realProgrammeMembershipId = UUID.fromString(id.toString());
-    } catch (IllegalArgumentException e) {
-      try {
-        curriculumMembershipId = (Long) id;
-      } catch (ClassCastException castException) {
-        throw new IllegalArgumentException(
-            String.format("%s is neither Long nor UUID", id));
-      }
-    }
-    if (realProgrammeMembershipId == null) {
-      LOG.info("Request received to save Conditions of Joining for Curriculum Membership {}.",
-          curriculumMembershipId);
-      programmeMembership = programmeMembershipService.findOne(curriculumMembershipId);
-    } else {
-      LOG.info("Request received to save Conditions of Joining for Programme Membership {}.",
-          realProgrammeMembershipId);
-      programmeMembership = programmeMembershipService.findOne(realProgrammeMembershipId);
+      programmeMembership = programmeMembershipService.findOne(Long.parseLong(id.toString()));
+    } catch (NumberFormatException e) {
+      programmeMembership = programmeMembershipService.findOne(UUID.fromString(id.toString()));
     }
 
     if (programmeMembership == null) {
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ConditionsOfJoiningServiceImpl.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ConditionsOfJoiningServiceImpl.java
@@ -53,7 +53,8 @@ public class ConditionsOfJoiningServiceImpl implements ConditionsOfJoiningServic
       try {
         curriculumMembershipId = (Long) id;
       } catch (ClassCastException castException) {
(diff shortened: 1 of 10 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ConditionsOfJoiningServiceImpl.java
```diff
@@ -43,16 +43,33 @@ public ConditionsOfJoiningServiceImpl(ConditionsOfJoiningRepository repository,
...
-        programmeMembershipId);
+  public ConditionsOfJoiningDto save(Object id, ConditionsOfJoiningDto dto) {
+    UUID realProgrammeMembershipId = null;
+    Long curriculumMembershipId = null;
+    ProgrammeMembershipDTO programmeMembership;
+    try {
+      realProgrammeMembershipId = UUID.fromString(id.toString());
+    } catch (IllegalArgumentException e) {
+      try {
+        curriculumMembershipId = (Long) id;
```
JosephKelly, line 54: My assumption would be that it's better to `Long.parseLong()` the `id.toString()` in case the id has been mapped as an Integer or other compatible number?

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ConditionsOfJoiningServiceImpl.java
```diff
@@ -43,16 +43,33 @@ public ConditionsOfJoiningServiceImpl(ConditionsOfJoiningRepository repository,
...
+    }
+    if (realProgrammeMembershipId == null) {
+      LOG.info("Request received to save Conditions of Joining for Curriculum Membership {}.",
+          curriculumMembershipId);
+      programmeMembership = programmeMembershipService.findOne(curriculumMembershipId);
+    } else {
+      LOG.info("Request received to save Conditions of Joining for Programme Membership {}.",
+          realProgrammeMembershipId);
+      programmeMembership = programmeMembershipService.findOne(realProgrammeMembershipId);
+    }
```
JosephKelly, line 68: It looks like could avoid the additional `if/else` and `realProgrammeMembershipId` by including the find methods in the `try/catch`, respectively.  Don't think I'll get a chance to test that refactor until much later.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/message/TraineeMessageListenerTest.java
```diff
@@ -29,25 +31,40 @@ void setUp() {
...
+    verify(service).save(CURRICULUM_MEMBERSHIP_ID, dto);
+  }
+
+  @Test
+  void shouldSaveSignedCojWhenEventValidWithPmId() {
+    ConditionsOfJoiningDto dto = new ConditionsOfJoiningDto();
+    dto.setSignedAt(SIGNED_AT);
+    dto.setVersion(GoldGuideVersion.GG9);
+    ConditionsOfJoiningSignedEvent event = new ConditionsOfJoiningSignedEvent(
+        PROGRAMME_MEMBERSHIP_ID, dto);
```
JosephKelly, line 52: While I'm nit-picking, this is now a duplicate block.  Following the pattern elsewhere, it would be moved to a "create" method.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/service/impl/ConditionsOfJoiningServiceImplTest.java
```diff
@@ -56,7 +56,31 @@ void saveShouldThrowExceptionWhenProgrammeMembershipIdNotFound() {
   }
 
   @Test
-  void saveShouldSaveTheConditionsOfJoiningAgainstTheProgrammeMembership() {
+  void saveShouldThrowExceptionWhenProgrammeMembershipIdNotFound() {
+    ConditionsOfJoiningDto coj = new ConditionsOfJoiningDto();
+    coj.setSignedAt(SIGNED_AT);
+    coj.setVersion(GoldGuideVersion.GG9);
```
JosephKelly, line 62: This could also could be extracted to a "create" method.

JosephKelly (changes requested):
The implementation comment is the one I'm most keen to see changed.  Casing to Long has some assumption baked in there.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ConditionsOfJoiningServiceImpl.java
ReubenRobertsHEE: Thanks, yes its a lot neater now!

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/message/TraineeMessageListenerTest.java
Judge40: Personally I lean towards duplication over `createXyz` methods for unit tests, unless we're also going to write a test for the createXyz method? :trollface: 

I find it easier to understand what the test is expecting when the initialisation is all there upfront. Otherwise I have to scroll around to see what the creation method is doing.

JosephKelly (approved):

PR #948, opened 2023-06-19
refactor:move delete curriculumMembership by id

hihilary:
to CurriculumMembershipService

TIS21-2403


JosephKelly (commented):
There are a few things in this.  I couldn't quickly find the comment that prompted this PR.  It looks part way toward having a parallel API -> Service -> Repository.
The 2 services are highly coupled so maybe shouldn't be separated if it means invoking back and forth.  Sorry if it was a bad idea.

Do let me know if you think this is a cleaner separation & I'll review more fully.  If so,it might just be 1 change.

PR #950, opened 2023-06-23
feat: skip auditing when no get method exists

JosephKelly:
This is a bug that follows modifying the method to get ProgrammeMembership by either a `Long` or `UUID` id.

We've got some divergence in how the service is structured. "CurriculumMemberships" are managed via a "ProgrammeMembershipResource". We haven't agreed how to properly fix this.

Taking a chance local sonar is more cautious than sonarcloud for NPE so might need to push something extra for that.

I started on a test but mocking the static method requires an upgrade of Mockito, which we get based on the version of spring.

NO-CARD: fixing deployment


JosephKelly:
I will raise this NPE bug as a false positive with Sonar.  It is specifically suggesting that `Class.getSimpleName()` could return a null value.

I also want to continue looking at the test I created, either activating the Spring Security Test features or whether we can use a version of Mockito that allows us to mock the `TisSecurityHelper.getProfileFromContext()`

hihilary (approved):

PR #953, opened 2023-06-28
fix:NPE for reval

hihilary:
fix NullPointerException for "/revalidation/trainee/{gmcId}" when gmcId not found in TCS.

TIS21-4548


jayanta2018 (approved):
LGTM

PR #954, opened 2023-06-29
fix:sending CurriculumMembershipDeletedEvent when curriculumMembership is deleted

hihilary:
In User Research-Revalidation channel, Katy raised a query on 27 June that they updated the programme membership info of a doctor but the doctor was still in the Discrepancy list.

This happened when the curriculum membership was deleted, and no message was sent to Reval to update the programme details.

no-ticket

### DIFF
--- a/tcs-service/pom.xml
+++ b/tcs-service/pom.xml
@@ -12,7 +12,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-service</artifactId>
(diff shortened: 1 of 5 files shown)


JosephKelly (approved):
Oh wow.
I think we should clean up the use of `ProgrammeMembership` in a `CurriculumMembershipDeletedEvent`.
Nice catch for the where we should be updating the elasticsearch document rather than deleting.

PR #955, opened 2023-07-03
fix:use PM.personId for ConnectionInfo

hihilary:
`personId` will no longer be in CurriculumMembership table.

We have 2 options to fix this in traineeInfoForConnection.sql:
1. just get rid of `WHERECLAUSE(cm, personId)` --> much lower the performance for a single record export (very quickly -> more than 2s)
2. join the ProgrammeMembership table to get personId --> lower the performance for the full sync (at most 2s more than before, and around 10s in total). I think this option is more acceptable.

TIS21-2403


JosephKelly (approved):
I'm approving because it looks ok & it is acceptable given:
1. You've clearly considered alternatives (thanks for including that in the PR)
1. I don't have the mental capacity to think about how we could improve the repeated use of the ProgrammeMembership table.

JosephKelly:
Also really pleased to see that you could change only the SQL & no java code :taco:

PR #962, opened 2023-07-25
feat: exclude F1 doctors from reval sync

CaiWillis:
TIS21-4459

### DIFF
--- a/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
+++ b/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
@@ -17,6 +17,8 @@ from (
     -- note: null values are filtered out by the condition below
     and lower(gmc.gmcNumber) <> 'unknown'
     and gmc.gmcNumber not like CONCAT('%', UNHEX('c2a0'), '%') -- filter out all gmc number with non-breaking space
+  inner join Placement pl on (pl.id = cd.id )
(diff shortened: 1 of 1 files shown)


hihilary (changes requested):
Got some questions to think about:
1. it's `Placement.traineeId` to check with `ContactDetails.id`
2. how do we handle null values (a person doesn't have any placements)
3. if a person has multiple placement, how we guarantee we export only 1 line

CaiWillis:
Will have a think and come back with some better ideas

tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
```diff
@@ -17,6 +17,8 @@ from (
     -- note: null values are filtered out by the condition below
     and lower(gmc.gmcNumber) <> 'unknown'
     and gmc.gmcNumber not like CONCAT('%', UNHEX('c2a0'), '%') -- filter out all gmc number with non-breaking space
+  inner join Placement pl on (pl.id = cd.id )
```
JosephKelly, line 20: Requires change: `Placement.id` should be replaced with `Placement.traineeId` and the scope limited to "current" placements.

If not already considered: please describe if doctors without a current placement are intentionally excluded.

JosephKelly: Too slow :turtle:

PR #964, opened 2023-07-27
fix:set null values when last ResultSet col was null and bump verison of tcs-client

hihilary:
1. bump version of tcs-client to lower the log level

2.  set null value when real value of getLong() was NULL, including:

-  PersonServiceImpl: mapper is used when `enableES` is false to decorate person views in the person list. As `enableES` is true in every environment, so I think this is not used, but it's a better-to-fix

- PlacementRowMapper: mapper is used in Peron details -> Placement Summary list. This fix will affect the UI:

Before (currently on Prod): 
image

After (tested locally):
image
As shown above, "0" site id will be removed from the placement summary list, leaving a pair of brackets there.

- PostViewRowMapper: mapper is used in PostView list. The fix will remove all "0" values from the parameters in the request to get the reference data, and it doesn't affect the UI.

- PostViewSearchMapper: only can be invoked via the endpoint "/api/findByNationalPostNumber", and the following decorator also shows the null values have to be removed: https://github.com/Health-Education-England/TIS-TCS/blob/10762eed0fcff7a44762b236858e9df15f3dd5cd/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/decorator/PostViewDecorator.java#L37

TIS21-4739

### DIFF
--- a/tcs-service/pom.xml
+++ b/tcs-service/pom.xml
@@ -12,7 +12,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-service</artifactId>
-  <version>6.31.2</version>
+  <version>6.31.3</version>
 
   <dependencies>
     <dependency>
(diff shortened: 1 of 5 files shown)


JosephKelly (approved):
I mean it feels a bit naughty to approve something I worked on but it's not raised by me :)

PR #966, opened 2023-07-28
feat: exclude doctors with a current F1 placement from the reval sync…

CaiWillis:
… and CDC

TIS21-4459

### DIFF
--- a/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
+++ b/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
@@ -8,13 +8,14 @@ from (
     pm1.programmeMembershipType,
     pm1.programmeStartDate,
     pm1.programmeEndDate,
+    placements.gradeAbbrvs,
     latestCm.curriculumEndDate,
     currentPmCounts.programmeNames as programmeName,
     if(currentPmCounts.count_num > 1, NULL, currentPmCounts.owner) as owner
   from
     ContactDetails cd
   inner join GmcDetails gmc on (gmc.id = cd.id)
-    -- note: null values are filtered out by the condition below
+   -- note: null values are filtered out by the condition below
     and lower(gmc.gmcNumber) <> 'unknown'
     and gmc.gmcNumber not like CONCAT('%', UNHEX('c2a0'), '%') -- filter out all gmc number with non-breaking space
   left join (
@@ -46,8 +47,18 @@ from (
     WHERECLAUSE(cm2, personId)
     group by cm2.programmeMembershipUuid
   ) latestCm on latestCm.programmeMembershipUuid = pm1.uuid
+  left join (
+      select
+  		pl.traineeId,
+  		-- one row per trainee
+  		GROUP_CONCAT(pl.gradeAbbreviation SEPARATOR " ") gradeAbbrvs
+        from Placement pl
+        where pl.dateFrom <= current_date() and pl.dateTo >= current_date()
+        group by pl.traineeId
+        ) placements
+          on placements.traineeId = cd.id
   WHERECLAUSE(cd, id)
-) as ot
+) as ot where ot.gradeAbbrvs is null or ot.gradeAbbrvs not LIKE '%F1%' -- include trainees without current placements and exclude f1
 ORDERBYCLAUSE
 LIMITCLAUSE
 ;
--- a/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
+++ b/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
@@ -54,7 +54,7 @@ from (
   		GROUP_CONCAT(pl.gradeAbbreviation SEPARATOR " ") gradeAbbrvs
         from Placement pl
         where pl.dateFrom <= current_date() and pl.dateTo >= current_date()
-        INDIVIDUALPLACEMENTCLAUSE
+        ANDWHERECLAUSE(pl, traineeId)
         group by pl.traineeId
         ) placements
           on placements.traineeId = cd.id
--- a/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
+++ b/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
@@ -15,7 +15,7 @@ from (
   from
     ContactDetails cd
   inner join GmcDetails gmc on (gmc.id = cd.id)
-   -- note: null values are filtered out by the condition below
+    -- note: null values are filtered out by the condition below
     and lower(gmc.gmcNumber) <> 'unknown'
     and gmc.gmcNumber not like CONCAT('%', UNHEX('c2a0'), '%') -- filter out all gmc number with non-breaking space
   left join (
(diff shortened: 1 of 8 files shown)


tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
```diff
@@ -46,8 +47,18 @@ from (
...
     group by cm2.programmeMembershipUuid
   ) latestCm on latestCm.programmeMembershipUuid = pm1.uuid
+  left join (
+      select
+  		pl.traineeId,
+  		-- one row per trainee
+  		GROUP_CONCAT(pl.gradeAbbreviation SEPARATOR " ") gradeAbbrvs
+        from Placement pl
+        where pl.dateFrom <= current_date() and pl.dateTo >= current_date()
+        group by pl.traineeId
```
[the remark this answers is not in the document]
CaiWillis: Will take a look. For the "Bulk Query", I'm getting 257828 rows returned in 20s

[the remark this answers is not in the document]
CaiWillis: Using the query as is but including the whereclauses I get a result in 11 seconds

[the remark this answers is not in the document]
CaiWillis: Have added an individual clause for placements and this reduces the time to < 1 second

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/RevalidationServiceImpl.java
```diff
@@ -228,10 +228,12 @@ public ConnectionInfoDto buildTcsConnectionInfo(Long personId) {
 
     // $1 and $2 will be injected with the parenthesized match sub patterns from the RegEx
     final String whereClause = String.format("where $1.$2 = %s", personId);
+    final String individualPlacementClause = String.format("and pl.traineeId = %s", personId);
 
     final String query = sqlQuerySupplier
         .getQuery(SqlQuerySupplier.TRAINEE_CONNECTION_INFO)
         .replaceAll(WHERE_CLAUSE_MACRO_REGEX, whereClause)
+        .replace("INDIVIDUALPLACEMENTCLAUSE", individualPlacementClause)
```
JosephKelly, line 236: The WHERECLAUSE was modified and made more flexible to accommodate field names & table aliases.
If we are going to stick with a separate template for this clause, I'll ask that this uses that same method to reduce the likelihood of the table alias `pl` and field name changing in one place but not the other.

JosephKelly (changes requested):
I think adding an extra clause like this makes it less consistent with the other persistence queries.
The proposed SQL does mean we have 1 less level of nesting but the consistency is better for me.

This commit is more consistent with the way that programme memberships are restricted: https://github.com/Health-Education-England/TIS-TCS/compare/main...feat/addCurrentPlacementGrade_TIS21-4769

I've left a couple of inline comments.

We also need to increment the version of the tcs-persistence module & use that in the `tcs-service` module.

tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
```diff
@@ -46,8 +47,19 @@ from (
...
+  		GROUP_CONCAT(pl.gradeAbbreviation SEPARATOR " ") gradeAbbrvs
+        from Placement pl
+        where pl.dateFrom <= current_date() and pl.dateTo >= current_date()
+        INDIVIDUALPLACEMENTCLAUSE
+        group by pl.traineeId
+        ) placements
+          on placements.traineeId = cd.id
   WHERECLAUSE(cd, id)
-) as ot
+) as ot where ot.gradeAbbrvs is null or ot.gradeAbbrvs not LIKE '%F1%' -- include trainees without current placements and exclude f1
```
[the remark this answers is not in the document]
CaiWillis: Would it work if I changed the group delimeter from a space to symbol (for argument's sake, | ) and did like %|F1|%?

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/RevalidationServiceImpl.java
CaiWillis: Cool, I could change it to an ANDWHERE clause

tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
```diff
@@ -46,8 +47,19 @@ from (
     WHERECLAUSE(cm2, personId)
     group by cm2.programmeMembershipUuid
   ) latestCm on latestCm.programmeMembershipUuid = pm1.uuid
+  left join (
+      select
+  		pl.traineeId,
+  		-- one row per trainee
+  		GROUP_CONCAT(pl.gradeAbbreviation SEPARATOR " ") gradeAbbrvs
```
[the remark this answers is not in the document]
CaiWillis: 🤦 thanks!

CaiWillis:
- Have removed the ANDWHERE clause and used the nested example
- Have added Regex and changes to group concat as suggested

Tested and the refactored query returns the same number of results as before, which is exatcly the number of F1 doctors less than the query before this commit, so all good

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/RevalidationServiceImpl.java
CaiWillis: Have used the nested query as suggested

tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
```diff
@@ -46,8 +47,21 @@ from (
...
   ) latestCm on latestCm.programmeMembershipUuid = pm1.uuid
+  left join (
+      select
+  		pl.traineeId,
+  		-- one row per trainee
+  		GROUP_CONCAT(distinct pl.gradeAbbreviation SEPARATOR " | ") currentGrades
+      from (
+        select traineeId, gradeAbbreviation, dateFrom, dateTo
+        from Placement pl
+        WHERECLAUSE(pl, id)
```
hihilary, line 58: I guess it's traineeId you want to filter here.

CaiWillis:
Note: change of approach, this PR is now intended to send the grade across to reval

tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
```diff
@@ -8,6 +8,7 @@ from (
     pm1.programmeMembershipType,
     pm1.programmeStartDate,
     pm1.programmeEndDate,
+    currentGrade.currentgrades,
```
hihilary, line 11: We need to use Camel Case too for the column name.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/RevalidationServiceImpl.java
```diff
@@ -66,14 +66,14 @@ public class RevalidationServiceImpl implements RevalidationService {
   private static final String PROGRAMME_NAME_FIELD = "programmeName";
   private static final String PROGRAMME_START_DATE_FIELD = "programmeStartDate";
   private static final String SURNAME_FIELD = "surname";
+  private static final String PLACEMENT_GRADE_FIELD = "currentgrades";
```
hihilary, line 69: The same here, camel case for column name.

hihilary (changes requested):

hihilary (approved):

PR #969, opened 2023-08-17
chore: increase TTL of user profile cache

JosephKelly:
This reduces the load on the profile service.
That is balanced with the need to expire removed users.

TIS21-4943: ESR processing affected by profile availability

### DIFF
--- a/.aws/tis-tcs-nimdta.json
+++ b/.aws/tis-tcs-nimdta.json
@@ -208,6 +208,10 @@
         {
           "name": "SENTRY_ENVIRONMENT",
           "value": "nimdta"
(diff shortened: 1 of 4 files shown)


CaiWillis (approved):

hihilary (approved):

JosephKelly:
Setting this environment variable overrides the default value specified in the profile client:  https://github.com/Health-Education-England/TIS-PROFILE/blob/414bdf41b73a9e3391c096aa2145b312edd21e0d/profile-client/src/main/resources/config/profileclientapplication.properties#L9

PR #974, opened 2023-08-30
fix: add getPlacementSummaryById

ReubenRobertsHEE:
Bugfix incompatibility between PlacementDetailsDto and PlacementSummaryDto expected by TIS-SYNC when using getPlacementById().

NO-TICKET

### DIFF
--- a/tcs-client/pom.xml
+++ b/tcs-client/pom.xml
@@ -6,7 +6,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-client</artifactId>
-  <version>6.2.0</version>
+  <version>6.2.1</version>
   <packaging>jar</packaging>
   <dependencies>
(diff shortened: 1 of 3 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PlacementService.java
```diff
@@ -58,6 +58,14 @@ public interface PlacementService {
    */
   PlacementDetailsDTO getDetails(Long id);
 
+  /**
+   * Get a list containing at most a single placement summary given the placement ID
+   *
+   * @param id the placement ID
+   * @return the placement summary if found
+   */
+  List<PlacementSummaryDTO> getSummary(Long id);
```
ReubenRobertsHEE, line 67: a "list" 😬

PR #975, opened 2023-09-05
fix:programmeMembership bulk update validate - case insensitive

hihilary:
Handle `rotation name`, `leaving reason` and `training pathway` when their values from bulk upload don't match the casing in Reference service or enumeration.

`ProgrammeMembershipType` is converted to an enumeration in Generic Upload service, so will need another PR there.

TIS21-3785


jayanta2018 (approved):
LGTM

PR #976, opened 2023-09-06
feat: filter dummy trainee role

jayanta2018:
Given that a doctor has a Person Role like ‘%Placeholder%’ or like ‘%Dummy%’
When Reval admin visits the discrepancies  page
Then they should NOT see that doctor

TIS21-4690

### DIFF
--- a/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
+++ b/tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
@@ -62,6 +62,12 @@ from (
     ) currentGrade on cd.id = currentGrade.traineeId
   WHERECLAUSE(cd, id)
 ) as ot
+-- Filter based on "role" from the "Person" table
+where ot.personId not in (
+  select distinct p.id
+  from Person p
+  where lower(p.role) like '%dummy%' or lower(p.role) like '%placeholder%'
+)
 ORDERBYCLAUSE
 LIMITCLAUSE
 ;


tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
```diff
@@ -62,6 +62,12 @@ from (
     ) currentGrade on cd.id = currentGrade.traineeId
   WHERECLAUSE(cd, id)
 ) as ot
+-- Filter based on "role" from the "Person" table
+WHERE ot.personId NOT IN (
+  SELECT DISTINCT p.id
+  FROM Person p
+  WHERE LOWER(p.role) LIKE '%dummy%' OR LOWER(p.role) LIKE '%placeholder%'
+)
```
CaiWillis, line 70: Can you put the sql operations into lowercase to match the rest of the file

This does appear to make the query perform slower, don't have any changes to suggest at this stage though

hihilary: We can try this on Stage. 

If there's a timeout or performance issue, we can get all the non-dummy records via jpa query, and then have a filter in Java.

tcs-persistence/src/main/resources/queries/traineeInfoForConnection.sql
```diff
@@ -18,6 +18,8 @@ from (
     -- note: null values are filtered out by the condition below
     and lower(gmc.gmcNumber) <> 'unknown'
     and gmc.gmcNumber not like CONCAT('%', UNHEX('c2a0'), '%') -- filter out all gmc number with non-breaking space
+  inner join Person p on (p.id = cd.id)
+    and LOWER(p.role) LIKE '%dummy%' OR LOWER(p.role) LIKE '%placeholder%'
```
hihilary, line 22: I think these can be removed.

jayanta2018: Oh i see my bad. Sorry I left it unintentionally

jayanta2018: Removed just now

hihilary (approved):

PR #979, opened 2023-09-15
fix: allow unknown properties for CoJ json

ReubenRobertsHEE:
To prevent errors such as
```
Caused by:
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "syncedAt" (class
com.transformuk.hee.tis.tcs.api.dto.ConditionsOfJoiningDto), not marked as ignorable (3 known properties: "version", "programmeMembershipUuid", "signedAt"])
```

`syncedAt` is used to record when the record is re-synced to TSS, it should always be null en-route to TIS.

TIS21-4714: Sync CoJ back from TIS to TSS

### DIFF
--- a/tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/mapper/ConditionsOfJoiningMapperTest.java
+++ b/tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/mapper/ConditionsOfJoiningMapperTest.java
@@ -1,0 +1,46 @@
+package com.transformuk.hee.tis.tcs.service.mapper;
+
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertNotNull;
+
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonMappingException;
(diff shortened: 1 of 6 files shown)


tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/mapper/ConditionsOfJoiningMapperTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Test
+  public void shouldIgnoreUnknownPropertiesWhenDeserializing() throws JsonProcessingException {
+
+    UUID uuid = UUID.randomUUID();
+    Instant signedAt = Instant.now();
+    GoldGuideVersion goldGuideVersion = GoldGuideVersion.GG9;
+
+    String jsonAsString =
+        "{\"programmeMembershipUuid\":\"" + uuid.toString() + "\"," +
```
Judge40, line 30: I was about to say how a text block would be more readable, then I remembered where we were 🙊

Judge40 (approved):
LGTM, technically should wait for @Health-Education-England/internal-admin-backend. But it's a tiny change, so I'll leave it up to you.

jayanta2018 (approved):

PR #985, opened 2024-01-11
feat: add fundingSubType field to PostFunding

CaiWillis:
TIS21-5381: Modify PostFunding Table to store FundingSubType uuid

### DIFF
--- a/tcs-api/pom.xml
+++ b/tcs-api/pom.xml
@@ -6,7 +6,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-api</artifactId>
-  <version>6.20.0</version>
+  <version>6.20.1</version>
   <packaging>jar</packaging>
 
   <properties>
--- a/tcs-api/pom.xml
+++ b/tcs-api/pom.xml
@@ -6,7 +6,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-api</artifactId>
-  <version>6.19.1</version>
+  <version>6.20.0</version>
   <packaging>jar</packaging>
 
   <properties>
--- a/tcs-api/pom.xml
+++ b/tcs-api/pom.xml
@@ -6,7 +6,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-api</artifactId>
-  <version>6.20.1</version>
+  <version>6.20.0</version>
   <packaging>jar</packaging>
 
   <properties>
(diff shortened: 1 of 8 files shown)


CaiWillis:
test coverage failure is on DTO, no other DTOs have test coverage, do we care?

tcs-api/src/main/java/com/transformuk/hee/tis/tcs/api/dto/PostFundingDTO.java
```diff
@@ -19,6 +19,8 @@ public class PostFundingDTO implements Serializable {
 
   private String fundingType;
 
+  private String fundingSubType;
```
hihilary, line 22: I think we ususally use UUID type for uuids, eg. CurrculumDTO, ConditionsOfJoiningDto, etc.

tcs-api/pom.xml
```diff
@@ -6,7 +6,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-api</artifactId>
-  <version>6.19.1</version>
+  <version>6.20.1</version>
```
hihilary, line 9: I think the next minor version is 6.20.0

tcs-persistence/pom.xml
```diff
@@ -13,12 +13,12 @@
   <description>A module containing the model and persistence layer for TCS</description>
   <groupId>uk.nhs.tis</groupId>
   <artifactId>tcs-persistence</artifactId>
-  <version>2.28.1</version>
+  <version>2.29.1</version>
```
hihilary, line 16: The same. The next minor version is 2.29.0

tcs-service/pom.xml
```diff
@@ -12,7 +12,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-service</artifactId>
-  <version>6.33.4</version>
+  <version>6.34.1</version>
```
hihilary, line 15: The same. The next minor version is 6.34.0

tcs-service/src/main/resources/db/migration/schema/V4.27__add_fundingsubtype_to_postfunding.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `PostFunding` ADD `fundingSubType` varchar(36);
```
hihilary, line 1: I'd prefer "ADD COLUMN" and it would be consistent with the other existing sql scripts.

tcs-service/pom.xml
CaiWillis: Ah I added the patches because I wasn't sure the analysis would work again, will revert

hihilary (approved):

PR #986, opened 2024-01-25
fix:postFundingMapper for PostResource

hihilary:
so frontend Post page will be able to create/update/show postfundings with fundingSubType

TIS21-5634


JosephKelly (approved):
I'll check the sonar definition of "changed code"... there are some method signatures that don't follow convention in the mapper!

PR #990, opened 2024-02-08
feat:add validation to PostFunding Update

hihilary:
The following PR is https://github.com/Health-Education-England/TIS-GENERIC-UPLOAD/pull/236

1. I couldn't find any clue the endpoints `/post-fundings/XXX` are used in the frontend UI.
2. In person validator, we separate the endpoint for bulk upload to a patch API, and for frontend endpoints we return a `MethodArgumentNotValidException`. But we didn't do this way in that way. I'm not very sure if I should create a new endpoin like what we did for person validator.

TIS21-5384

### DIFF
--- a/tcs-service/pom.xml
+++ b/tcs-service/pom.xml
@@ -12,7 +12,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-service</artifactId>
-  <version>6.35.0</version>
+  <version>6.36.0</version>
 
   <dependencies>
     <dependency>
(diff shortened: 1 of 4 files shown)


JosephKelly (approved):
Just confirming this doesn't validate the subtype.
That's good for my assumptions about how this method will be used but I wanted to check there weren't uncommitted changes to the validator.

hihilary:
> Just confirming this doesn't validate the subtype. That's good for my assumptions about how this method will be used but I wanted to check there weren't uncommitted changes to the validator.

It validates the subtype, in `validateFundingType()` method, there's a call of `checkFundingSubType()`.

PR #991, opened 2024-02-14
Feat/coj into pm dto v3squashed jpa joins

JosephKelly:
I was going to do the refactoring based on the //TODO in the code but there is a more significant refactoring to be done, removing references to the ID received and simplifying the call stack invoked from the `@RabbitListener`.


tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/ConditionsOfJoining.java
```diff
@@ -15,12 +20,17 @@
...
   @Type(type = "org.hibernate.type.UUIDCharType")
   private UUID programmeMembershipUuid;
   private Instant signedAt;
   @Enumerated(EnumType.STRING)
   private GoldGuideVersion version;
+
+  @OneToOne(mappedBy = "conditionsOfJoining", cascade = CascadeType.REMOVE)
+  @PrimaryKeyJoinColumn(name = "programmeMembershipUuid")
+  @JoinColumn(name = "programmeMembershipUuid")
+  private ProgrammeMembership programmeMembership;
```
hihilary, line 35: `programmeMembershipUuid` above looks like a duplicate field now.

JosephKelly: It's a bit weird because it's both the foreign key but also the PK.

hihilary (approved):

PR #993, opened 2024-02-20
fix: catch additional exceptions

JosephKelly:
This is very belt and braces because of it shouldn't have happened. We should not have ended up with a  DataAccessEXception. We catch that but also check the returned ProgrammeMembership.

NO-CARD: Catch-up time


Judge40 (approved):

hihilary (approved):

JosephKelly:
Soz... I didn't want to merge it with that double space

hihilary (approved):

PR #996, opened 2024-03-15
feat: add fundingStatus field to Post and initialise values with flyway

CaiWillis:
TIS21-5808: Add a new field fundingStatus in Post


CaiWillis:
Unsure about tests, it's possible what I've done is a bit of a fudge just to get things passing

Low test coverage appears to be existing "issue" - not sure we usually care about entity and DTO getters and setters

hihilary (approved):

PR #1005, opened 2024-07-29
fix: CoJ find-PM code refactor

ReubenRobertsHEE:
To fix issue with missing PM causing infinite message processing as described
https://hee-nhs-tis.slack.com/archives/C03G3UK3U82/p1722243933162119

(There are other places in the code that use getOne() but I think in a safe manner because they immediately retrieve the object.)

NO-TICKET

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ConditionsOfJoiningServiceImpl.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ConditionsOfJoiningServiceImpl.java
@@ -56,28 +56,24 @@ public class ConditionsOfJoiningServiceImpl implements ConditionsOfJoiningServic
   public ConditionsOfJoiningDto save(Object id, ConditionsOfJoiningDto dto) {
     Optional<ProgrammeMembership> programmeMembershipOptional = Optional.empty();
     LOG.info("Request received to save Conditions of Joining for id {}.", id);
-    try {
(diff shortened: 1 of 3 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ConditionsOfJoiningServiceImpl.java
```diff
@@ -52,24 +54,32 @@ public ConditionsOfJoiningServiceImpl(ConditionsOfJoiningRepository repository,
...
+        programmeMembershipOptional = programmeMembershipRepository.findByUuid(
+            UUID.fromString(id.toString()));
+      }
+
+      if (!programmeMembershipOptional.isPresent()) {
         throw new IllegalArgumentException(String.format("No Programme Membership for %s.", id));
       }
-      LOG.debug("Found ProgrammeMembership {}", programmeMembership);
+      LOG.debug("Found ProgrammeMembership {}", programmeMembershipOptional.get());
     } catch (EntityNotFoundException e) {
```
Judge40, line 77: Can this still be thrown?

ReubenRobertsHEE: Ah no 🤦  hang on let me remove that.

Judge40 (approved):

JosephKelly (approved):

PR #1007, opened 2024-08-13
fix:expire postEsrEvent when tcs receives an event

hihilary:
I raised a [draft PR](https://github.com/Health-Education-England/TIS-EsrReconciliationService/pull/244) in EsrReconciliationService, but due to Java version & SonarCloud incompability, that PR was not able to get analysed, so I parked that one, and tried to resolve the issue on TCS. 
However, in TCS, it's not able to have the info if the current event will expire any other events from ESR, so we have to check the db table every time.

1. for the flywayscript, below are the result run locally on my machine:
```
2024-08-12 09:58:35.528  INFO 23376 --- [  restartedMain] o.f.core.internal.command.DbMigrate      : Migrating schema `tcs` to version 4.33 - update expired postEsrEvent
2024-08-12 10:00:28.870  INFO 23376 --- [  restartedMain] o.f.core.internal.command.DbMigrate      : Successfully applied 1 migration to schema `tcs` (execution time 01:53.374s)
```
2. the flyway script will only keep the latest reconciled record as RECONCILED and set all other previously reconciled records to RECONCILED_EXP.

TIS21-6306


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
```diff
@@ -668,11 +670,51 @@ public Optional<PostEsrEventDto> markPostAsEsrPositionChanged(
...
+    Long positionNumber = postEsrExportedDto.getPositionNumber();
+    Long currentPostId = postEsrExportedDto.getPostId();
+    LocalDateTime currentReconciledTime = postEsrExportedDto.getEventDateTime();
+    boolean isCurrentEventExpired = false;
+    Set<PostEsrEvent> postEsrEvents = postEsrEventRepository.findPostEsrEventsByPositionNumber(
+        positionNumber);
+
+    Set<PostEsrEvent> expiredPostEsrEvents = postEsrEvents.stream().filter(
+        postEsrEvent -> postEsrEvent.getStatus() == PostEsrEventStatus.RECONCILED
+            && !postEsrEvent.getPost().getId().equals(currentPostId)
```
hihilary, line 699: I'm not sure if this condidion is required as we can set all previous reconciled records to EXP.

tcs-api/src/main/java/com/transformuk/hee/tis/tcs/api/enumeration/PostEsrEventStatus.java
```diff
@@ -6,5 +6,6 @@
 public enum PostEsrEventStatus {
 
   RECONCILED,
-  DELETED
+  DELETED,
+  RECONCILED_EXP
```
JosephKelly, line 10: Based on conversations, I believe that `EXP` is short for expired rather than an unusual bit of ESR terminology like APC.
Would `EXPIRED` or `EXPIRED_RECONCILIATION` be a suitable name?  I also like `INVALIDATED` slightly more than `OBSOLETE` as a more common way of saying `DEPRICATED`.  If `REPLACED` isn't too confusing, that would be more consistent with "plain english".

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
```diff
@@ -668,11 +670,51 @@ public Optional<PostEsrEventDto> markPostAsEsrPositionChanged(
         .postEsrEventDtoToPostEsrEvent(postEsrExportedDto);
     newPostEsrEvent.setPost(post);
 
+    boolean isCurrentEventExpired = handPositionReconciliationExpiry(postEsrExportedDto);
+    if (isCurrentEventExpired && newPostEsrEvent.getStatus() == PostEsrEventStatus.RECONCILED) {
+      newPostEsrEvent.setStatus(PostEsrEventStatus.RECONCILED_EXP);
+    }
```
JosephKelly, line 676: I found I was jumping back and forth to understand and review the change.

JosephKelly (commented):
For something to be an `Event`, I don't think we would typically modify/re-write it.  Changing the existing event would be inconsistent with having something that was reconciled and later deleted.

We do have constraints caused by the data model & having multiple records of the same event and a very broad `findBy...()`.  I probably need a bit more context over why we want to update records in this way.

We can look to reduce exacerbating duplicates by modifying the concurrency/rate of the caller.

hihilary:
After discussion, we will set a view for only latest ESR events for every position number.

PR #1012, opened 2024-08-16
fix: remove training number persistence

Judge40:
The generated training number should not be persisted as the training number database table is legacy.

TIS21-6375


jayanta2018 (dismissed):

jayanta2018:
looks like build failed

hihilary (approved):

PR #1016, opened 2024-09-05
fix:the programme filter when the user is assigned programmes

hihilary:
fix the programme filter when the user is assigned programmes, and refactor `searchForPage` method to remove duplicate codes.

1. added the columnFilter to `RoleBasedFilterStrategy.getFilter`, because when we apply the programmeId filter, we need to check if the programmeMembershipStatus filter is also applied. And currently, on the UI, we can only select either CURRENT or FUTURE, not both of them.
2. resume a `general programmeMembership.status filter` in `searchForPage` method in case no programmeId or programmeName filters are applied
3. tested locally under both scenarios if the user is assgined any programmes or not. I tried to set up some unit tests, but it asked for token in the context. For integration tests, we haven't properly set up an ES, so I finally gave up.

TIS21-6359

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/strategy/ProgrammeBasedFilterStrategy.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/strategy/ProgrammeBasedFilterStrategy.java
@@ -3,33 +3,80 @@ package com.transformuk.hee.tis.tcs.service.strategy;
 import com.transformuk.hee.tis.security.model.Programme;
 import com.transformuk.hee.tis.security.model.UserProfile;
 import com.transformuk.hee.tis.security.util.TisSecurityHelper;
+import com.transformuk.hee.tis.tcs.api.enumeration.ProgrammeMembershipStatus;
+import com.transformuk.hee.tis.tcs.service.model.ColumnFilter;
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.lucene.search.join.ScoreMode;
 import org.elasticsearch.common.collect.Tuple;
 import org.elasticsearch.index.query.BoolQueryBuilder;
 import org.elasticsearch.index.query.MatchQueryBuilder;
 import org.elasticsearch.index.query.NestedQueryBuilder;
+import org.elasticsearch.index.query.QueryBuilders;
 import org.springframework.stereotype.Component;
 
 @Component
 public class ProgrammeBasedFilterStrategy implements RoleBasedFilterStrategy {
 
-  private static final String COLUMN_FILTER = "programmeName";
+  private static final String COLUMN_FILTER = "programmeId";
 
   @Override
-  public Optional<Tuple<String, BoolQueryBuilder>> getFilter() {
+  public Optional<Tuple<String, BoolQueryBuilder>> getFilter(List<ColumnFilter> columnFilters) {
+    MatchQueryBuilder statusMatchQueryBuilder = null;
+    if (columnFilters != null) {
+      Set<ColumnFilter> statusFilters = columnFilters.stream().filter(
+              columnFilter -> StringUtils.equals(columnFilter.getName(), "programmeMembershipStatus"))
+          .collect(
+              Collectors.toSet());
+      if (statusFilters.size() > 0) {
+        ProgrammeMembershipStatus status = ProgrammeMembershipStatus
+            .valueOf(statusFilters.iterator().next().getValues().get(0).toString());
+        statusMatchQueryBuilder = getProgrammeMembershipStatusQueryBuilder(status);
+      }
+    }
+
     UserProfile currentUserProfile = TisSecurityHelper.getProfileFromContext();
     Set<Programme> assignedProgrammes = currentUserProfile.getAssignedProgrammes();
     if (CollectionUtils.isNotEmpty(assignedProgrammes)) {
       BoolQueryBuilder programmeRoleFilter = new BoolQueryBuilder();
-      assignedProgrammes.forEach(programme -> programmeRoleFilter.should(
-          new NestedQueryBuilder("programmeMemberships",
-              new MatchQueryBuilder("programmeMemberships.programmeId", programme.getId()),
-              ScoreMode.None)));
+      for (Programme programme : assignedProgrammes) {
+        if (statusMatchQueryBuilder == null) {
+          programmeRoleFilter.should(
+              new NestedQueryBuilder("programmeMemberships",
+                  new MatchQueryBuilder("programmeMemberships.programmeId", programme.getId()),
+                  ScoreMode.None));
+        } else {
+          programmeRoleFilter.should(
+              new NestedQueryBuilder("programmeMemberships",
+                  new BoolQueryBuilder().should(
+                          new MatchQueryBuilder("programmeMemberships.programmeId", programme.getId()))
+                      .minimumShouldMatch(1)
+                      .should(statusMatchQueryBuilder).minimumShouldMatch(2),
+                  ScoreMode.None));
+        }
+      }
       return Optional.of(new Tuple<>(COLUMN_FILTER, programmeRoleFilter));
     }
     return Optional.empty();
+  }
+
+  public MatchQueryBuilder getProgrammeMembershipStatusQueryBuilder(
+      ProgrammeMembershipStatus status) {
+    MatchQueryBuilder statusQueryBuilder = null;
+    if (status.equals(ProgrammeMembershipStatus.CURRENT)) {
+      statusQueryBuilder = QueryBuilders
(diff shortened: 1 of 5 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/strategy/ProgrammeBasedFilterStrategy.java
```diff
@@ -3,33 +3,80 @@
...
 @Component
 public class ProgrammeBasedFilterStrategy implements RoleBasedFilterStrategy {
 
-  private static final String COLUMN_FILTER = "programmeName";
+  private static final String COLUMN_FILTER = "programmeId";
 
   @Override
-  public Optional<Tuple<String, BoolQueryBuilder>> getFilter() {
+  public Optional<Tuple<String, BoolQueryBuilder>> getFilter(List<ColumnFilter> columnFilters) {
+    MatchQueryBuilder statusMatchQueryBuilder = null;
```
JosephKelly, line 29: It looks like the `statusMatchQueryBuilder` is only used `if (CollectionUtils.isNotEmpty(assignedProgrammes))` so could sit within that block.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/strategy/ProgrammeBasedFilterStrategy.java
```diff
@@ -3,33 +3,80 @@
...
-      assignedProgrammes.forEach(programme -> programmeRoleFilter.should(
-          new NestedQueryBuilder("programmeMemberships",
-              new MatchQueryBuilder("programmeMemberships.programmeId", programme.getId()),
-              ScoreMode.None)));
+      for (Programme programme : assignedProgrammes) {
+        if (statusMatchQueryBuilder == null) {
+          programmeRoleFilter.should(
+              new NestedQueryBuilder("programmeMemberships",
+                  new MatchQueryBuilder("programmeMemberships.programmeId", programme.getId()),
+                  ScoreMode.None));
```
JosephKelly, line 51: There is some duplication here that might improve maintainability here by making the intention clearer.  I'm happy enough if you want to leave this.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/strategy/ProgrammeBasedFilterStrategy.java
```diff
@@ -3,33 +3,80 @@
...
+        } else {
+          programmeRoleFilter.should(
+              new NestedQueryBuilder("programmeMemberships",
+                  new BoolQueryBuilder().should(
+                          new MatchQueryBuilder("programmeMemberships.programmeId", programme.getId()))
+                      .minimumShouldMatch(1)
```
JosephKelly, line 57: I think this is redundant, being replaced on the next line with `minimumShouldMatch(2)`.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/strategy/ProgrammeBasedFilterStrategy.java
```diff
@@ -3,33 +3,80 @@
...
   @Override
-  public Optional<Tuple<String, BoolQueryBuilder>> getFilter() {
+  public Optional<Tuple<String, BoolQueryBuilder>> getFilter(List<ColumnFilter> columnFilters) {
+    MatchQueryBuilder statusMatchQueryBuilder = null;
+    if (columnFilters != null) {
+      Set<ColumnFilter> statusFilters = columnFilters.stream().filter(
+              columnFilter -> StringUtils.equals(columnFilter.getName(), "programmeMembershipStatus"))
+          .collect(
+              Collectors.toSet());
+      if (statusFilters.size() > 0) {
```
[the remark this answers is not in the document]
JosephKelly: If you wanted to get really funky, you could use a `Stream.map` to return an `Optional<MatchQueryBuilder> statusMatchQueryBuilder`.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/strategy/ProgrammeBasedFilterStrategy.java
```diff
@@ -3,33 +3,80 @@
...
 import org.elasticsearch.index.query.MatchQueryBuilder;
 import org.elasticsearch.index.query.NestedQueryBuilder;
+import org.elasticsearch.index.query.QueryBuilders;
 import org.springframework.stereotype.Component;
 
 @Component
 public class ProgrammeBasedFilterStrategy implements RoleBasedFilterStrategy {
 
-  private static final String COLUMN_FILTER = "programmeName";
+  private static final String COLUMN_FILTER = "programmeId";
```
JosephKelly, line 25: :+1:  This is ok given the rest of the code & my assumption it was `programmeName` to restrict user-selected filtering.  It would be useful to look at whether we can clear up some of the nesting & logic associated with avoiding multiple sets of filtering.  The previous commit for this line reads like a squash so it is hard to say.

I don't know why this was there despite having a "must" across filters. It is a bit of an awkward one.  I expect it was `programmeName` to restrict user-selected filtering.  It would be useful to look at whether we can clear up some of the nesting & logic associated with avoiding multiple sets of filtering.

JosephKelly (changes requested):
This seems like a great refactor & with some local testing to work through, saw how the filtering was happening.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/strategy/ProgrammeBasedFilterStrategy.java
```diff
@@ -3,33 +3,80 @@
...
+          .matchQuery("programmeMemberships.programmeMembershipStatus", "CURRENT");
+    } else if (status.equals(ProgrammeMembershipStatus.PAST)) {
+      statusQueryBuilder = QueryBuilders
+          .matchQuery("programmeMemberships.programmeMembershipStatus", "PAST");
+    } else if (status.equals(ProgrammeMembershipStatus.FUTURE)) {
+      statusQueryBuilder = QueryBuilders
+          .matchQuery("programmeMemberships.programmeMembershipStatus", "FUTURE");
+    }
+    return statusQueryBuilder;
+  }
```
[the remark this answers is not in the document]
hihilary: It was copy and paste. Thanks for the refactoring.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/strategy/ProgrammeBasedFilterStrategy.java
hihilary: Done.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/strategy/ProgrammeBasedFilterStrategy.java
hihilary: I prefer to leaving as what it is.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/strategy/ProgrammeBasedFilterStrategy.java
hihilary: Removed. Nice spot.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/strategy/ProgrammeBasedFilterStrategy.java
hihilary: Yeah, I assumed it was `programmeName` for user-selected filtering too.

I'm not sure if I understand this comment properly. But the "must" filter makes sense to me as I think in this "must" filter, every filtering is required. 
This `programmeMembershipStatus` filter is very tricky. It can be used on its own, or on the user-selected `programmeName` column filter or role based `programmeId` filter. I found it was really hard to clean up any other bit of filtering.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/strategy/ProgrammeBasedFilterStrategy.java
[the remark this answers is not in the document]
hihilary: > N.B. As there is no handling for multiple filters of the same column, that would accept potentially let deterministic behaviour if there 2 filters of the same name were passed in.

I think this is a general problem for all APIs working with column filters and has to be handled in the same way.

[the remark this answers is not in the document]
JosephKelly: the N.B. should read

> ...potentially **less** deterministic...

i.e. the suggesstion to use a `Stream` might interpret the parameter differently across invocations because a Stream doesn't have the ordering constraints of a List.  I'm happy with that.

JosephKelly (approved):

PR #1018, opened 2024-09-09
fix: Add COJ Gold Guide 10

DorisWongNHS:


tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/service/impl/ConditionsOfJoiningServiceImplTest.java
```diff
@@ -134,6 +134,27 @@ void saveShouldSaveTheConditionsOfJoiningAgainstTheProgrammeMembershipFromPmId()
         is(GoldGuideVersion.GG9));
   }
 
+  @Test
```
Judge40, line 137: This could be combined with the above test by using a parameterised test with enum source

JosephKelly (approved):

PR #1019, opened 2024-09-11
feat: update GMC details when provided by trainee

Judge40:
Update a trainee's GMC details when a GmcDetailsProvided event is received.
Perform validation on both the event and updated GMC details to ensure that the new details are valid.

TIS21-6292
TIS21-6479


JosephKelly (approved):
The separate `personId` seems redundant.
Happy with it as is though.

Judge40:
> The separate `personId` seems redundant. Happy with it as is though.

It was either that or not using the existing GmcDetailsDto. The structure of the event is sensible considering its from an external system, it's TIS business logic/data structure that states the gmc details include a person ID

PR #1020, opened 2024-09-16
fix:remove the outer programmeMembership status filter

hihilary:
Previously this filter didn't work, and last iteration we added it back, which resulted in some users who don't have programme admin/observers roles not able to see person records which don't have any active programme memberships. So we decided to take it off as a patch to resume this functionality to be the same as before.

However, the programme membership filter doesn't work well and has to be reviewed in the future.

Tested locally.

TIS21-6359

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PersonElasticSearchService.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PersonElasticSearchService.java
@@ -136,10 +136,6 @@ public class PersonElasticSearchService {
               continue;
             }
             if (StringUtils.equals(columnFilter.getName(), "programmeMembershipStatus")) {
-              shouldBetweenSameColumnFilter.should(new NestedQueryBuilder("programmeMemberships",
-                  new MatchQueryBuilder("programmeMemberships.programmeMembershipStatus",
-                      value.toString()),
-                  ScoreMode.None));
               continue;
             }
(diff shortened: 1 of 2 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PersonElasticSearchService.java
```diff
@@ -136,10 +136,6 @@ public Page<PersonViewDTO> searchForPage(String searchQuery,
               continue;
             }
             if (StringUtils.equals(columnFilter.getName(), "programmeMembershipStatus")) {
-              shouldBetweenSameColumnFilter.should(new NestedQueryBuilder("programmeMemberships",
-                  new MatchQueryBuilder("programmeMemberships.programmeMembershipStatus",
-                      value.toString()),
-                  ScoreMode.None));
               continue;
```
JosephKelly, line 139: Given most requests end up including a programme membership status, it would be better to add this to the check on [L#121-122](https://github.com/Health-Education-England/TIS-TCS/pull/1020/files#diff-49f81cca9561db7d6db7c513b2f40a4ba94ba3da3ecb2263f7e272f86faba8a6L121-L122).

JosephKelly (changes requested):
I think this improves the situation for most users but might still be overly restrictive for some users.  I'll create a follow up story looking at:
- What a Programme User should be able to filter by (across programme name & status)
- Whether it is appropriate to fix everyone's search of programme name to just "current" ones
- The clients we control (UI)

JosephKelly (approved):

PR #1022, opened 2024-09-16
fix: use machine user to call reference

Judge40:
The GMC Validator checks the given GMC status against the reference service to check if the value is valid. However, calls to the erference service use the current user's authentication token. As the new GMC update is received via RabbitMQ there is no authenticated user to provide a token.

Update ReferenceClientProdConfig to use a machine user token instead of requiring an actual user token.

TIS21-6292
TIS21-6479


JosephKelly (approved):
With no guarantees about preserving this funcionality, I'm happy for this to be applied

PR #1024, opened 2024-09-24
add an endpoint to recieve a query string of programme membership ids…

EdwardBarclay:
… and return an array of programmeMembershipSummaryDto for each id

PR https://github.com/Health-Education-England/TIS-ADMINS-UI/pull/1414 is dependent on this PR being in place before merging.

TIS21-6276

### DIFF
--- a/tcs-api/src/main/java/com/transformuk/hee/tis/tcs/api/dto/ProgrammeMembershipSummaryDTO.java
+++ b/tcs-api/src/main/java/com/transformuk/hee/tis/tcs/api/dto/ProgrammeMembershipSummaryDTO.java
@@ -1,0 +1,15 @@
+package com.transformuk.hee.tis.tcs.api.dto;
+
+import java.io.Serializable;
+import java.time.LocalDate;
+import lombok.Data;
+
+/**
+ * A DTO for the ProgrammeMembershipSummary entity.
+ */
+@Data
+public class ProgrammeMembershipSummaryDTO implements Serializable {
+  private String programmeMembershipId;
+  private String programmeMembershipName;
+  private LocalDate programmeStartDate;
+}
--- a/tcs-api/src/main/java/com/transformuk/hee/tis/tcs/api/dto/ProgrammeMembershipSummaryDTO.java
+++ b/tcs-api/src/main/java/com/transformuk/hee/tis/tcs/api/dto/ProgrammeMembershipSummaryDTO.java
@@ -9,7 +9,7 @@ import lombok.Data;
  */
 @Data
 public class ProgrammeMembershipSummaryDTO implements Serializable {
-  private String programmeMembershipId;
-  private String programmeMembershipName;
+  private String programmeMembershipUuid;
+  private String programmeName;
   private LocalDate programmeStartDate;
 }
--- a/tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/repository/ProgrammeMembershipRepository.java
+++ b/tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/repository/ProgrammeMembershipRepository.java
@@ -43,15 +43,6 @@ public interface ProgrammeMembershipRepository extends JpaRepository<ProgrammeMe
       + "WHERE pm.programme.id = :programmeId")
   List<ProgrammeMembership> findByProgrammeId(@Param("programmeId") Long programmeId);
 
-  /**
-   * Get a list of programme memberships.
-   *
-   * @param uuids List of all the programme memberships.
-   * @return a list of programme memberships for all the uuid.
-   */
-  List<ProgrammeMembership> findProgrammeMembershipsByUuidIn(
-      List<UUID> uuids);
-
   List<ProgrammeMembership> findByUuidIn(Set<UUID> ids);
 
   Optional<ProgrammeMembership> findByUuid(UUID id);
--- a/tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/repository/ProgrammeMembershipRepository.java
+++ b/tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/repository/ProgrammeMembershipRepository.java
@@ -43,6 +43,15 @@ public interface ProgrammeMembershipRepository extends JpaRepository<ProgrammeMe
       + "WHERE pm.programme.id = :programmeId")
   List<ProgrammeMembership> findByProgrammeId(@Param("programmeId") Long programmeId);
 
+  /**
+   * Get a list of programme memberships.
+   *
+   * @param uuids List of all the programme memberships.
+   * @return a list of programme memberships for all the uuid.
+   */
+  List<ProgrammeMembership> findProgrammeMembershipsByUuidIn(
+      List<UUID> uuids);
+
   List<ProgrammeMembership> findByUuidIn(Set<UUID> ids);
 
   Optional<ProgrammeMembership> findByUuid(UUID id);
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/ProgrammeMembershipResource.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/ProgrammeMembershipResource.java
@@ -3,6 +3,7 @@ package com.transformuk.hee.tis.tcs.service.api;
 import com.transformuk.hee.tis.tcs.api.dto.CurriculumMembershipDTO;
 import com.transformuk.hee.tis.tcs.api.dto.ProgrammeMembershipCurriculaDTO;
 import com.transformuk.hee.tis.tcs.api.dto.ProgrammeMembershipDTO;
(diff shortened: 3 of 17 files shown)


JosephKelly (dismissed):
This absolutely needs some tests, TDD might help you work through it.  An integration test should give you what you need without the need for a unit test as well.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/ProgrammeMembershipResource.java
```diff
@@ -160,6 +160,52 @@ public ResponseEntity<ProgrammeMembershipDTO> getProgrammeMembership(@PathVariab
...
+      @PathVariable String ids) {
+
+    String[] idArray = ids.split(",");
+    List<ProgrammeMembershipSummaryDto> summaryList = new ArrayList<>();
+
+    for (String id : idArray) {
+      ProgrammeMembershipDTO programmeMembershipDto = null;
+
+      try {
+        programmeMembershipDto = programmeMembershipService.findOne(UUID.fromString(id));
```
JosephKelly, line 182: I forgot to say, I think there's probably a more appropriate method on the service to use for this search rather than looping over each id separately to do a `findOne`.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/ProgrammeMembershipResource.java
```diff
@@ -160,6 +160,52 @@ public ResponseEntity<ProgrammeMembershipDTO> getProgrammeMembership(@PathVariab
...
+   * @param ids a comma-separated list of programmeMembershipDTO ids to retrieve
+   * @return the ResponseEntity with status 200 (OK) and with body the programmeMembershipDTO list,
+   *         or with status 404 (Not Found)
+   */
+  @GetMapping("/programme-memberships/{ids}/summary-list")
+  @PreAuthorize("hasPermission('tis:people::person:', 'View')")
+  public ResponseEntity<List<ProgrammeMembershipSummaryDto>> getProgrammeMembershipSummaryList(
+      @PathVariable String ids) {
+
+    String[] idArray = ids.split(",");
```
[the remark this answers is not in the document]
EdwardBarclay: I am updating it to be more inline for REST principles using query parameters which are more appropriate for multiple paramaters

EdwardBarclay:
> This absolutely needs some tests, TDD might help you work through it. An integration test should give you what you need without the need for a unit test as well.

Test was delayed because the project is not allowing me to test claiming missing queues and flayway files, definitely adding a test TDD would have been preferable

tcs-api/src/main/java/com/transformuk/hee/tis/tcs/api/dto/ProgrammeMembershipSummaryDTO.java
```diff
@@ -0,0 +1,15 @@
...
+import java.time.LocalDate;
+import lombok.Data;
+
+/**
+ * A DTO for the ProgrammeMembershipSummary entity.
+ */
+@Data
+public class ProgrammeMembershipSummaryDTO implements Serializable {
+  private String programmeMembershipId;
+  private String programmeMembershipName;
```
ReubenRobertsHEE, line 13: Do we mean 'programmeName' here? Otherwise I'm not sure what this field would contain, so it should have a comment if its not a std. ProgrammeMembership field I think.

tcs-api/src/main/java/com/transformuk/hee/tis/tcs/api/dto/ProgrammeMembershipSummaryDTO.java
```diff
@@ -0,0 +1,15 @@
...
+import java.io.Serializable;
+import java.time.LocalDate;
+import lombok.Data;
+
+/**
+ * A DTO for the ProgrammeMembershipSummary entity.
+ */
+@Data
+public class ProgrammeMembershipSummaryDTO implements Serializable {
+  private String programmeMembershipId;
```
ReubenRobertsHEE, line 12: Not sure what field is meant here: is it the uuid? If so, it should probably be called 'uuid' (the 'id' field in the ProgrammeMembership entity is deprecated).

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/repository/ProgrammeMembershipRepository.java
```diff
@@ -43,6 +43,15 @@ List<ProgrammeMembership> findAllProgrammeMembershipInDescOrderByTraineeId(
...
   List<ProgrammeMembership> findByProgrammeId(@Param("programmeId") Long programmeId);
 
+  /**
+   * Get a list of programme memberships.
+   *
+   * @param uuids List of all the programme memberships.
+   * @return a list of programme memberships for all the uuid.
+   */
+  List<ProgrammeMembership> findProgrammeMembershipsByUuidIn(
+      List<UUID> uuids);
```
ReubenRobertsHEE, line 53: Given there is already something which takes a `Set<UUID>` and returns the same content, is it not possible to reuse that? It feels a little awkward to duplicate the same functionality here.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/ProgrammeMembershipResource.java
```diff
@@ -148,7 +170,6 @@ public ResponseEntity<List<ProgrammeMembershipDTO>> getAllProgrammeMemberships(
   @GetMapping("/programme-memberships/{id}")
   @PreAuthorize("hasPermission('tis:people::person:', 'View')")
   public ResponseEntity<ProgrammeMembershipDTO> getProgrammeMembership(@PathVariable String id) {
```
ReubenRobertsHEE, line 172: Unless there is a good reason for the change, probably best to leave the debug logging 😁

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/ProgrammeMembershipResource.java
```diff
@@ -61,6 +67,21 @@ public ProgrammeMembershipResource(ProgrammeMembershipService programmeMembershi
     this.programmeMembershipValidator = programmeMembershipValidator;
   }
 
+  /**
+   * This mapper is used to convert the detailed programme membership data into a summary form
+   * by extracting specific fields such as the programme name and start date.
+   */
+  @Mapper
```
ReubenRobertsHEE, line 74: Is there a reason why this isn't put into the service.mappers package? It feels a bit odd to have it included here.

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/repository/ProgrammeMembershipRepository.java
EdwardBarclay: the reason I have it as a List alternative is because we can receive duplicate ids such as if a form A and B are for the same programme, a list allows duplicates where as a Set is unique only

ReubenRobertsHEE: Ah, ok. Do you have another PR that I can use to see the whole journey? It still feels odd to me to need to include duplicates, since the ProgrammeMembershipSummaryDTO content will be identical for these, so I would have thought that a single entry could be 'duplicated' as necessary when producing the forms listing, but maybe I'm misunderstanding the process?

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/ProgrammeMembershipResource.java
```diff
@@ -160,6 +165,33 @@ public ResponseEntity<ProgrammeMembershipDTO> getProgrammeMembership(@PathVariab
...
+   */
+  @GetMapping("/programme-memberships/summary-list")
+  @PreAuthorize("hasPermission('tis:people::person:', 'View')")
+  public ResponseEntity<List<ProgrammeMembershipSummaryDTO>> getProgrammeMembershipSummaryList(
+      @RequestParam List<String> ids) {
+    Set<UUID> uuids = ids.stream()
+        .map(UUID::fromString)
+        .collect(Collectors.toSet());
+    List<ProgrammeMembershipDTO> programmeMembershipDtos = programmeMembershipService
+        .findProgrammeMembershipsByUuid(uuids);
```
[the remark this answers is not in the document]
ReubenRobertsHEE: Hi Ed, this looks good! Do you think you could implement a test or two to demonstrate what happens when (some of) the UUIDs are malformed, or don't match to anything in the DB? Do we get nothing, or a partial listing excluding those bad parameters?

ReubenRobertsHEE (approved):

ReubenRobertsHEE (dismissed):
Do get approval from the admin team before merging please :)

ReubenRobertsHEE (approved):

PR #1027, opened 2024-10-11
fix: use existing repository method

ReubenRobertsHEE:
I think this could work - and it saves you having to set up a ProgrammeMembershipRepository test for your new method ;)


EdwardBarclay (approved):
I will tweak the call from front end to avoid duplicates

PR #1030, opened 2024-10-23
refactor: create a filter to ensure that only the valid uuids are mapped

EdwardBarclay:


Judge40 (changes requested):
You could just write this in the filter instead of adding an extra service method for something quite simple/generic.

```java
Set<UUID> uuids = ids.stream()
    .filter(id -> {
      try {
        UUID.fromString(id);
        return true;
      } catch (IllegalArgumentException e) {
        return false;
      }
    })
    .map(UUID::fromString)
    .collect(Collectors.toSet());
```

ReubenRobertsHEE:
We'll also need a test for this, one way or another.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/ProgrammeMembershipResourceIntTest.java
```diff
@@ -313,24 +314,24 @@ void shouldGetProgrammeMembershipSummaryListByUuid2() throws Exception {
     curriculumMembership.setCurriculumEndDate(today.plusDays(1));
     programmeMembershipRepository.saveAndFlush(programmeMembership);
 
-    UUID savedProgrammeMembershipUuid = programmeMembership.getUuid();
+    String ProgrammeMembershipUuid = String.valueOf(programmeMembership.getUuid()) + ",77777";
+    String FilterdUuidList = String.valueOf(programmeMembership.getUuid());
```
Judge40, line 318: Variable names should be lower camel case, can also just use `programmeMembership.getUuid().toString()` instead of `String.valueOf()` since we know it's not going to be null.

Judge40 (approved):

PR #1032, opened 2024-10-30
a quick fix for tcs logic as the response was causing the forms to be…

EdwardBarclay:
… prevented from loading


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/ProgrammeMembershipResource.java
```diff
@@ -197,7 +195,7 @@ public ResponseEntity<List<ProgrammeMembershipSummaryDTO>> getProgrammeMembershi
         summaryList = programmeMembershipService
             .findProgrammeMembershipSummariesByUuid(uuids);
         if (summaryList.isEmpty()) {
-          return ResponseEntity.status(HttpStatus.NOT_FOUND).body(Collections.emptyList());
+          return ResponseEntity.ok(summaryList);
         }
         return ResponseEntity.ok(summaryList);
```
Judge40, line 200: Both return values are the same, so you can remove the condition

Judge40 (changes requested):

PR #1035, opened 2024-12-06
feat: send rejected GMC updates to tis-sync

ReubenRobertsHEE:
For the sync service to pass on to kinesis and tis-trainee-sync. Related PR: https://github.com/Health-Education-England/TIS-SYNC/pull/256

TIS21-6639

### DIFF
--- a/tcs-service/pom.xml
+++ b/tcs-service/pom.xml
@@ -324,6 +324,12 @@
       <artifactId>sentry-logback</artifactId>
       <version>${sentryVersion}</version>
     </dependency>
+    <dependency>
+      <groupId>com.amazonaws</groupId>
+      <artifactId>amazon-sqs-java-messaging-lib</artifactId>
+      <version>1.0.8</version>
+      <type>jar</type>
+    </dependency>
 
   </dependencies>
   <packaging>war</packaging>
--- a/tcs-service/pom.xml
+++ b/tcs-service/pom.xml
@@ -12,7 +12,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-service</artifactId>
-  <version>6.47.5</version>
+  <version>6.48.0</version>
   <dependencies>
     <dependency>
       <groupId>org.springframework.boot</groupId>
--- a/.aws/tis-tcs-nimdta.json
+++ b/.aws/tis-tcs-nimdta.json
@@ -182,6 +182,10 @@
           "valueFrom": "trainee-rabbit-gmc-details-provided-queue-tcs-nimdta"
         },
         {
+          "name": "AWS_SQS_SYNC_QUEUE_URL",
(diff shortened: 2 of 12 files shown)


tcs-service/pom.xml
```diff
@@ -324,6 +324,12 @@
       <artifactId>sentry-logback</artifactId>
       <version>${sentryVersion}</version>
     </dependency>
+    <dependency>
+      <groupId>com.amazonaws</groupId>
+      <artifactId>amazon-sqs-java-messaging-lib</artifactId>
+      <version>1.0.8</version>
+      <type>jar</type>
+    </dependency>
```
JosephKelly, line 332: How essential is this?  Dependency & memory bloat is already an issue for TCS.

TCS already uses Rabbit MQ & there is a trainee service that acts as an intermediary between rabbit & SQS.
I'd probably want to understand the trade-offs before reviewing deeper.

ReubenRobertsHEE: Hi Pepe. I agree it's not great to have to add another messaging library to TCS. The reason I implemented via SQS is that the tis-sync service already receives requests for data via SQS (e.g. https://github.com/Health-Education-England/TIS-SYNC/blob/df9e5cc8f8cc334d3cb239b1be04f7a3d81284ab/src/main/java/uk/nhs/tis/sync/job/RecordResendingJob.java#L32 ) so it seemed to fit within that paradigm. 

If we were to rather use rabbit it would require more changes to tis-sync to accept data requests in that manner as well as via SQS, or a rework of upstream services to send requests with rabbit. We could do this, but it felt like a more significant set of changes than something to embark on simply as an 'implementation detail' of the ticket.

JosephKelly: I think a call/something to look at the architecture would be useful.  I did assume this was a 'least-worse' option rather than architectural change by design.

IMHO, we own the system architecture so rather than being bound to work within it or the limitations of how it works & growing associated Tech Debt, we [~can modify~ | have the responsibility to] update how the whole-system works by evolving the architecture.

My head's a bit full but I've noticed that the sync service includes AMQP config & processing already.  It might be too complicated, but failed updates are currently written to a dead letter queue so creating routing for Exceptions would be something to consider, along with 

The consideration is heavily influenced by the ownership of the service, in production.
I am thinking of the coupling between services pattern & this being more of an asynchronous "push" rather than adding some consequential "resync" (reaction) to an event-driven "Failed Update" (thing that happens and published, currently to a dead letter queue).

ReubenRobertsHEE: Thanks Pepe. Some of this is quite subtle, so I will need to mull it over. I am getting the suggestion to look at picking up the message from the DLQ in tis-sync, and hence not make any changes to tis-tcs at all. I'll see if I can make any headway with a reasonable implementation of that, seems like a good option.

PR #1037, opened 2025-01-08
chore: Define max JVM memory

JosephKelly:
Tasks appear to be hitting a ceiling at 42% of RAM available. This has means the earlier memory increase is less effective. Adding the maximum size improves the ability to autoscale on memory.

TIS21-6834: Optimise JVM memory use

### DIFF
--- a/tcs-service/Dockerfile
+++ b/tcs-service/Dockerfile
@@ -4,4 +4,4 @@ EXPOSE 8093
 
 COPY target/app.jar app.jar
 
-CMD ["java", "-jar", "app.jar"]
+CMD ["java", "-XX:MaxRAMPercentage=95", "-jar", "app.jar"]
--- a/tcs-service/Dockerfile
+++ b/tcs-service/Dockerfile
@@ -4,4 +4,4 @@ EXPOSE 8093
 
 COPY target/app.jar app.jar
 
-CMD ["java", "-XX:MaxRAMPercentage=95", "-jar", "app.jar"]
+CMD ["java", "-XX:MaxRAMPercentage=95.0", "-jar", "app.jar"]


tobi-nhs (approved):

tcs-service/Dockerfile
```diff
@@ -4,4 +4,4 @@ EXPOSE 8093
 
 COPY target/app.jar app.jar
 
-CMD ["java", "-jar", "app.jar"]
+CMD ["java", "-XX:MaxRAMPercentage=95", "-jar", "app.jar"]
```
CaiWillis, line 7: My understanding of this is pretty poor, but as I understand this affects the heap size, does 95% leave enough for other processes? (Guess it depends on how grabby it is, i.e." just because it can have 95 doesn't mean it gets 95")

tobi-nhs: 95% is allowed usable memory

jayanta2018, line 7: Looks good to me. Just wondering, together with this, can we also use garbage collector  "-XX:+UseG1GC" ?

JosephKelly: Valid consideration I had only given a little thought to.

- This is the only container in each task
- Within the container there are kernel processes etc. that will need to consume some memory.  We may need to further tune the amount but want to make sure we have a level that can be used to indicate a need for additional instances of the task to be run.

JosephKelly: I did mean to note a larger concern around the use of these containers outside of ECS, i.e. local environments.  In those situations, there is a greater chance of exhausting system resources.  Some but not all ways of running the container will allow defining a limit to the memory available.

That ought to be a consideration of the separate story we have to look at the container runtime we use for running dev/local environments.

JosephKelly: We absolutely could look to add additional optimisations into the JVM.  I want to avoid over-polishing work for the small well defined story this satisfies.

CaiWillis (approved):

jayanta2018 (approved):

PR #1040, opened 2025-01-10
feat: get GMC by id

ReubenRobertsHEE:
Since findGmcDetailsIn() actually uses GMC numbers to search by, not ids. tis-sync PR to use this new client function to follow.

TIS21-6639

### DIFF
--- a/tcs-client/src/main/java/com/transformuk/hee/tis/tcs/client/service/impl/TcsServiceImpl.java
+++ b/tcs-client/src/main/java/com/transformuk/hee/tis/tcs/client/service/impl/TcsServiceImpl.java
@@ -698,6 +698,12 @@ public class TcsServiceImpl extends AbstractClientService {
     return responseEntity.getBody();
   }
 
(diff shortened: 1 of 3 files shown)


tcs-client/src/main/java/com/transformuk/hee/tis/tcs/client/service/impl/TcsServiceImpl.java
```diff
@@ -698,6 +698,12 @@ public List<GmcDetailsDTO> findGmcDetailsIn(List<String> gmcIds) {
     return responseEntity.getBody();
   }
 
+  public GmcDetailsDTO getGmcDetailsById(Long id) {
```
ReubenRobertsHEE, line 701: I didn't include a javadoc because there aren't any in this file. I can amend if needed, but then perhaps we need to write javadocs for all the client functions?

hihilary (approved):

PR #1047, opened 2025-02-06
feat: make post funding required for post create and update, make fun…

CaiWillis:
…ding-type mandatory field

TIS21-6914: Make Post Funding mandatory for all Current Posts

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PostValidator.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PostValidator.java
@@ -306,6 +306,10 @@ public class PostValidator {
             && !postFundingDto.getEndDate().isAfter(postFundingDto.getStartDate())) {
           fieldErrors.add(new FieldError(POST_DTO_NAME, "fundings",
               "Post funding end date must not be equal to or before start date"));
+        } else if (postFundingDto.getFundingType() == null) {
+          fieldErrors.add(
+              new FieldError(POST_DTO_NAME, "fundings",
+              "Post Funding must have a funding type"));
         }
       }
     }
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PostValidator.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PostValidator.java
@@ -43,6 +43,7 @@ public class PostValidator {
   public static final String POST_DTO_NAME = "PostDTO";
   public static final String SPECIALTIES = "specialties";
   public static final String NATIONAL_POST_NUMBER = "nationalPostNumber";
+  public static final String FUNDINGS = "fundings";
   private ProgrammeRepository programmeRepository;
   private PostRepository postRepository;
   private SpecialtyRepository specialtyRepository;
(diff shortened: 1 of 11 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PostValidator.java
```diff
@@ -306,8 +307,16 @@ private Collection<? extends FieldError> checkFunding(PostDTO postDto) {
...
+        } else if (postFundingDto.getFundingType() == null) {
+          fieldErrors.add(
+              new FieldError(POST_DTO_NAME, "fundings",
+              "Post Funding must have a funding type"));
         }
       }
+    } else if (postDto.getStatus() == Status.CURRENT) { //No Post Funding but CURRENT
+      fieldErrors.add(
+          new FieldError(POST_DTO_NAME, "fundings",
+              "Current Posts Must have at least 1 Post Funding"));
```
hihilary, line 319: PostDto.status is not required and is populated from post fundings, so I think backend should not rely on the status field from the client.

CaiWillis: Do you have any advice on a good approach for this? It feels odd to be fetching from the DB at this point but I guess that's the only option given that the field is a backend-only managed one?

hihilary: I think the `validate` method is only used by createPost & updatePost methods in PostResource.
And the client should send all the post fundings attached to the post to the backend. So if you want to check if there's one postFunding, you can just check the dto directly.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResourceIntTest.java
```diff
@@ -288,27 +311,13 @@ public void initTest() {
     postGrade = createPostGrade(GRADE_ID, PostGradeType.APPROVED, post);
     postSite = createPostSite(SITE_ID, PostSiteType.PRIMARY, post);
     postSpecialty = createPostSpecialty(specialty, PostSpecialtyType.PRIMARY, post);
+    postFundings = createPostFundings(post);
+    postFundings.stream().forEach(pf -> em.persist(pf));
```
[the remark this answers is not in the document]
CaiWillis: This turned out the be that cause of most of my issues, thanks!

hihilary (changes requested):
1. Do we need to check if the postFunding Set is an empty set?
2. I think we need more tests in PostResourceIntTest to verify the behaviours if postFunding is null (or empty?).

CaiWillis:
> 1. Do we need to check if the postFunding Set is an empty set?
> 2. I think we need more tests in PostResourceIntTest to verify the behaviours if postFunding is null (or empty?).

1. Good catch, I've replaced the annotation with NotEmpty and added integration tests
2. Done https://github.com/Health-Education-England/TIS-TCS/pull/1047/commits/2f5e68377deb3ec88f4d15904ee86b94a62d5a94

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PostValidator.java
```diff
@@ -306,6 +306,10 @@ private Collection<? extends FieldError> checkFunding(PostDTO postDto) {
             && !postFundingDto.getEndDate().isAfter(postFundingDto.getStartDate())) {
           fieldErrors.add(new FieldError(POST_DTO_NAME, "fundings",
               "Post funding end date must not be equal to or before start date"));
+        } else if (postFundingDto.getFundingType() == null) {
+          fieldErrors.add(
+              new FieldError(POST_DTO_NAME, "fundings",
+              "Post Funding must have a funding type"));
```
hihilary, line 312: I'm a little wondering if this validation in PostValidator is required as Spring takes it over to validate postFunding NotEmpty in the dto.

CaiWillis: Sorry are you referring to line 300 where we check for empty set? We still need to check each individual post funding for a funding type right?

hihilary: Yes, I did misunderstood the ACs.

hihilary (approved):

PR #1052, opened 2025-03-13
feat: delete post

jayanta2018:
TIS21-7021

- Create a check on the User's Profile on Designated Body so that only those where their Designated Body matches the Post Owner has access to the delete function on a post.
- Create a new role called 'Post Management' and assign the post delete permission to this role (Profile service)
- Posts with placements will not be able to be deleted

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/PostResource.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/PostResource.java
@@ -309,7 +309,7 @@ public class PostResource {
    * @return the ResponseEntity with status 200 (OK)
    */
   @DeleteMapping("/posts/{id}")
-  @PreAuthorize("hasAuthority('post:delete:posts')")
+  @PreAuthorize("hasAuthority('post:delete')")
   public ResponseEntity<Void> deletePost(@PathVariable Long id) {
     log.debug("REST request to delete Post : {}", id);
     postService.checkTheLoggedInUserDbSameAsPostOwner(id, getProfileFromContext());
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/PostResource.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/PostResource.java
@@ -309,7 +309,7 @@ public class PostResource {
    * @return the ResponseEntity with status 200 (OK)
    */
   @DeleteMapping("/posts/{id}")
-  @PreAuthorize("hasAuthority('tcs:delete:entities') and hasAuthority('post:delete:posts')")
+  @PreAuthorize("hasAuthority('post:delete:posts')")
   public ResponseEntity<Void> deletePost(@PathVariable Long id) {
     log.debug("REST request to delete Post : {}", id);
     postService.checkTheLoggedInUserDbSameAsPostOwner(id, getProfileFromContext());
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/PostResource.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/PostResource.java
@@ -1,5 +1,6 @@
 package com.transformuk.hee.tis.tcs.service.api;
 
+import static com.transformuk.hee.tis.security.util.TisSecurityHelper.getProfileFromContext;
 import static uk.nhs.tis.StringConverter.getConverter;
 
 import com.google.common.collect.Lists;
@@ -308,11 +309,10 @@ public class PostResource {
    * @return the ResponseEntity with status 200 (OK)
    */
   @DeleteMapping("/posts/{id}")
-  @PreAuthorize("hasAuthority('tcs:delete:entities')")
+  @PreAuthorize("hasAuthority('tcs:delete:entities') and hasAuthority('post:delete:posts')")
   public ResponseEntity<Void> deletePost(@PathVariable Long id) {
     log.debug("REST request to delete Post : {}", id);
-    postService.canLoggedInUserViewOrAmend(id);
-
+    postService.checkTheLoggedInUserDbSameAsPostOwner(id, getProfileFromContext());
     postService.delete(id);
     return ResponseEntity.ok()
         .headers(HeaderUtil.createEntityDeletionAlert(ENTITY_NAME, id.toString())).build();
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/PostResource.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/PostResource.java
@@ -1,6 +1,5 @@
 package com.transformuk.hee.tis.tcs.service.api;
 
-import static com.transformuk.hee.tis.security.util.TisSecurityHelper.getProfileFromContext;
 import static uk.nhs.tis.StringConverter.getConverter;
 
 import com.google.common.collect.Lists;
@@ -312,12 +311,10 @@ public class PostResource {
   @PreAuthorize("hasAuthority('post:delete')")
   public ResponseEntity<Void> deletePost(@PathVariable Long id) {
     log.debug("REST request to delete Post : {}", id);
-    postService.checkTheLoggedInUserDbSameAsPostOwner(id, getProfileFromContext());
     postService.delete(id);
     return ResponseEntity.ok()
         .headers(HeaderUtil.createEntityDeletionAlert(ENTITY_NAME, id.toString())).build();
   }
-
 
   /**
    * POST  /bulk-posts : Bulk create a new Posts.
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostService.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostService.java
@@ -180,7 +180,7 @@ public interface PostService {
 
   /**
    * Method that will throw a not authorized exception if the current logged in user's
-   * designated body does not match with post owner
+   * designated body does not match with post owner.
    *
    * @param postId the db managed id of the post record
    * @param userProfile logged in user's profile
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostService.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostService.java
@@ -174,9 +174,9 @@ public interface PostService {
    * Method that will throw a not authorized exception if the current logged in user cannot view or
    * modify the person record
    *
-   * @param personId the db managed id of the person record
+   * @param postId the db managed id of the post record
    */
-  void canLoggedInUserViewOrAmend(Long personId);
+  void canLoggedInUserViewOrAmend(Long postId);
 
   /**
    * Method that will throw a not authorized exception if the current logged in user's
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostService.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostService.java
@@ -1,6 +1,5 @@
 package com.transformuk.hee.tis.tcs.service.service;
 
-import com.transformuk.hee.tis.security.model.UserProfile;
 import com.transformuk.hee.tis.tcs.api.dto.PostDTO;
 import com.transformuk.hee.tis.tcs.api.dto.PostEsrDTO;
 import com.transformuk.hee.tis.tcs.api.dto.PostEsrEventDto;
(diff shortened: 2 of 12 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
```diff
@@ -664,7 +666,11 @@ public PostDTO findOne(Long id) {
   @Override
   public void delete(Long id) {
     log.debug("Request to delete Post : {}", id);
-    postRepository.deleteById(id);
+    if(permissionService.canDeletePostData()) {
+      postRepository.deleteById(id);
+      List<PostFunding> postFundingsToDelete = postFundingRepository.findByPostId(id);
+      postFundingRepository.deleteAll(postFundingsToDelete);
+    }
```
hihilary, line 673: If the users don't have the permission `post:delete:entities`, they won't get any error.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PermissionService.java
```diff
@@ -22,6 +22,7 @@ public class PermissionService {
   protected static final String VIEW_SENSITIVE_DATA_ETL_ROLE = "ETL";
   protected static final String VIEW_SENSITIVE_DATA_ROLE = "personsensitive:view:entities";
   protected static final String EDIT_SENSITIVE_DATA_ROLE = "personsensitive:add:modify:entities";
+  protected static final String DELETE_POST_DATA_ROLE = "post:delete:entities"; // should it be "post:delete:posts"?
```
hihilary, line 25: I don't quite understand what the relationship between the 2 permissions: `post:delete:entities` and `tcs:delete:entities`

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
jayanta2018: I have moved it to hasAuthority

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PermissionService.java
jayanta2018: I have changed it

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
```diff
@@ -664,7 +669,14 @@ public PostDTO findOne(Long id) {
   @Override
   public void delete(Long id) {
     log.debug("Request to delete Post : {}", id);
-    postRepository.deleteById(id);
+    List<Placement> attachedPlacements = placementRepository.findByPostId(id);
+    if (attachedPlacements.isEmpty()) {
+      postRepository.deleteById(id);
+      List<PostFunding> postFundingsToDelete = postFundingRepository.findByPostId(id);
+      postFundingRepository.deleteAll(postFundingsToDelete);
```
JosephKelly, line 676: Getting the post would mean that JPA joined placements & funding can be fetched and/or deleted without having to do that ourselves and including another dependency (code smell).  Please which is preferable.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/PostResource.java
```diff
@@ -308,11 +309,10 @@ public ResponseEntity<List<PlacementSummaryDTO>> getPlacementsForPosts(
    * @return the ResponseEntity with status 200 (OK)
    */
   @DeleteMapping("/posts/{id}")
-  @PreAuthorize("hasAuthority('tcs:delete:entities')")
+  @PreAuthorize("hasAuthority('post:delete:posts')")
```
jayanta2018, line 312: I have tested locally with the profile service. With necessary changes in the profile DB and code it works fine. Refer to profile PR

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
```diff
@@ -664,7 +666,16 @@ public PostDTO findOne(Long id) {
   @Override
   public void delete(Long id) {
     log.debug("Request to delete Post : {}", id);
-    postRepository.deleteById(id);
+    Optional<Post> optionalPost = postRepository.findById(id);
+    if (optionalPost.isPresent()) {
+      Set<Placement> attachedPlacements = optionalPost.get().getPlacementHistory();
+      if (attachedPlacements.isEmpty()) {
+        postRepository.deleteById(id);
+        // Related PostFunding records will be cascade deleted
```
jayanta2018, line 674: It has also been tested in the local.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostService.java
```diff
@@ -173,9 +174,18 @@ Page<PostViewDTO> findByNationalPostNumber(String query, List<ColumnFilter> colu
    * Method that will throw a not authorized exception if the current logged in user cannot view or
    * modify the person record
    *
-   * @param personId the db managed id of the person record
+   * @param postId the db managed id of the post record
    */
-  void canLoggedInUserViewOrAmend(Long personId);
+  void canLoggedInUserViewOrAmend(Long postId);
```
JosephKelly, line 179: :slightly_smiling_face:

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
JosephKelly: Comment seems unnecessary.  The same is true of all attached entities marked with the appropriate annotations.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
```diff
@@ -699,6 +710,19 @@ public void canLoggedInUserViewOrAmend(Long postId) {
...
 
+  @Override
+  public void checkTheLoggedInUserDbSameAsPostOwner(Long postId, UserProfile userProfile) {
+    Optional<Post> optionalPost = postRepository.findById(postId);
+    if (optionalPost.isPresent()) {
+      String postOwner = optionalPost.get().getOwner();
+      if (!DesignatedBodyMapper.
+          map(userProfile.getDesignatedBodyCodes()).contains(postOwner)) {
+        throw new AccessUnauthorisedException(
+            "You cannot delete Post with id: " + postId);
```
JosephKelly, line 721: This can be a single line.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResource2Test.java
```diff
@@ -315,21 +314,6 @@ public void getPostPlacementShouldReturnUnauthWhenUserNotPartOfSameTrustNew() th
...
-
-    doThrow(new AccessUnauthorisedException("")).when(postService)
-        .canLoggedInUserViewOrAmend(personId);
-
-    restPostMockMvc.perform(delete("/api/posts/{id}", personId)
-        .contentType(MediaType.APPLICATION_JSON))
-        .andExpect(status().isUnauthorized())
-        .andExpect(jsonPath("$.message").value("error.accessDenied"));
-    verify(postService, never()).delete(any());
-  }
```
JosephKelly, line 331: An equivalent test is still useful for the LocalOffice restrictions.
Please make sure you are adequately covering the ACs.  I've checked a few files for a test but don't see an existing one.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
```diff
@@ -699,6 +710,19 @@ public void canLoggedInUserViewOrAmend(Long postId) {
...
+      }
+    }
+  }
```
JosephKelly, line 724: If this is unlikely to be called anywhere other than for deleting, as indicated by the Exception message, we can avoid repeated calls to fetch the Post by including it in the delete operation.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostService.java
```diff
@@ -173,9 +174,18 @@ Page<PostViewDTO> findByNationalPostNumber(String query, List<ColumnFilter> colu
...
+
+  /**
+   * Method that will throw a not authorized exception if the current logged in user's
+   * designated body does not match with post owner.
+   *
+   * @param postId the db managed id of the post record
+   * @param userProfile logged in user's profile
+   */
+  void checkTheLoggedInUserDbSameAsPostOwner(Long postId, UserProfile userProfile);
```
JosephKelly, line 188: May be unneeded as per comment on the implementation class.

JosephKelly (commented):
I've lost at least one of the things I was looking at last time so will give this a further review after the permission is checked.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostService.java
jayanta2018: Yeah, I also thought so but when I call it from PostResource using the "postService" reference then it fails. Interface PostService needs this method here so that PostServiceImpl class can provide the implementation detail.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
jayanta2018: Refer to the previous comment in the PostResource interface

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/PostResource.java
jayanta2018: Changed

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
```diff
@@ -666,12 +666,12 @@ public PostDTO findOne(Long id) {
   @Override
   public void delete(Long id) {
     log.debug("Request to delete Post : {}", id);
+    checkIfReconciledWithESR(id);
```
JosephKelly: I'm fairly sure this will need to be changed & done differently.
Please discuss this, using checks / warnings elsewhere in the system, where we display something to users before they modify data.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/PostResource2Test.java
```diff
@@ -314,6 +315,20 @@ public void getPostPlacementShouldReturnUnauthWhenUserNotPartOfSameTrustNew() th
     verify(placementSummaryDecorator, never()).decorate(any());
   }
 
+  @Test
+  public void deletePostShouldReturnUnauthWhenUserDbNotSameAsPostOwner() throws Exception {
+    long postId = 1L;
+
+    doThrow(new AccessUnauthorisedException("")).when(postService)
+        .checkTheLoggedInUserDbSameAsPostOwner(eq(postId), any(UserProfile.class));
```
JosephKelly, line 323: If you accept `any(UserProfile.class)`, then you wouldn't need to invoke `mockUserprofile()` for this test.  Is it done for some other tests?  If I remember correctly, you only needed to add one for an integration test.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImplTest.java
```diff
@@ -443,6 +440,21 @@ void deleteShouldDeleteOneInstanceById() {
     verify(postRepositoryMock).deleteById(1L);
   }
 
+  @Test
+  void deleteShouldFailWhenPostIsReconciledWithESR() {
```
JosephKelly: Please confirm this is true.
Please also confirm the expected behaviour for the reconciliation events (should deletes cascade), if a delete should succeed when post is reconciled with ESR.

JosephKelly (changes requested):
The code smell is a significant one which may lead to unexpected behaviour. See the "Why is this an issue?" description.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImplTest.java
jayanta2018: Yes it does

JosephKelly (dismissed):

JosephKelly (approved):

PR #1056, opened 2025-05-01
fix: change specialty type fetch to eager

jayanta2018:
When I was  trying to delete a Post, which is linked to a Specialty, and Specialty has:

@ElementCollection(targetClass = SpecialtyType.class, fetch = FetchType.LAZY)

AuditingAspect tries to access the specialtyTypes set during auditing, Hibernate tries to lazy-load the collection, but the session is already closed or unavailable which causes this error:

"failed to lazily initialize a collection of role: Specialty.specialtyTypes, could not initialize proxy - no Session"

So the FetchType is changed to EAGER. In the DB we have less than 1200 records so performance will not be an issue.

----------------------------------------------------------------------------------------------------------------------------------------------------------------
Audit logs were prohibiting the post delete to go forward by checking the SpecialtyType, which was on FetchType Lazy loading that caused the Hibernate session to expire. (See above)

Introducing an AuditHelperService class and ensuring this helper is annotated @Transactional so that lazy fields are fetched safely. 

Annotate the audit aspect or audit call with @Transactional(propagation = Propagation.REQUIRES_NEW) so that it opens a new Hibernate session for loading lazy fields.

TIS21-7021

### DIFF
--- a/tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/Specialty.java
+++ b/tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/model/Specialty.java
@@ -48,7 +48,7 @@ public class Specialty implements Serializable {
   @Column(name = "specialtyCode")
   private String specialtyCode;
 
-  @ElementCollection(targetClass = SpecialtyType.class, fetch = FetchType.EAGER)
+  @ElementCollection(targetClass = SpecialtyType.class, fetch = FetchType.LAZY)
   @CollectionTable(name = "SpecialtyTypes", joinColumns = {@JoinColumn(name = "specialtyId")})
   @Column(name = "specialtyType")
   @Enumerated(EnumType.STRING)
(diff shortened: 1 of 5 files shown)


JosephKelly (commented):
I've tried to run locally to verify this approach.
I still see the error `failed to lazily initialize a collection of role: com.transformuk.hee.tis.tcs.service.model.Specialty.specialtyTypes, could not initialize proxy - no Session` which no longer prevents a deletion.

Local debugging suggests that this is a problem reading the nested entity:
1. newPost or oldPost
2. specialties[*]
3. specialtyType

That debugging showed that unlike other entities that are audited prior to delete (e.g. placement), attempting to inspect the response entity & dto gives a LazyInitializationException, indicating they are being proxied by Hibernate.

I am not marking this as `Request changes` because I'm unable to sufficiently understand the approach.  If someone else understands this well enough to verify it works as intended, that's good enough for me.

jayanta2018:
Solution does not lie here.

PR #1057, opened 2025-05-04
fix: fix audit aspect hibernate session for specialty type

jayanta2018:
Aspect Oriented Programming: Auditing Aspect's hibernate session expiry was found eventually at the mappers, i.e. PostMapper and SpecialtyMapper where specialty dtos and entities setting specialty types. result.setSpecialtyTypes()

TIS21-7021

### DIFF
--- a/tcs-client/pom.xml
+++ b/tcs-client/pom.xml
@@ -6,7 +6,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-client</artifactId>
-  <version>6.7.1</version>
+  <version>6.7.2</version>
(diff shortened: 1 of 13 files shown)


tcs-client/pom.xml
```diff
@@ -6,7 +6,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-client</artifactId>
-  <version>6.7.1</version>
+  <version>6.7.2</version>
```
jayanta2018, line 9: Risk here: other services using tcs-client needs to be updated. i.e. generic upload, ESR?

PR #1059, opened 2025-05-27
Feat/add validation mode tis21 7284

hihilary:
For any dto updates, don't validate fields if there're no changes.

TIS21-7284

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/GdcDetailsResource.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/GdcDetailsResource.java
@@ -69,7 +69,7 @@ public class GdcDetailsResource {
       @RequestBody @Validated(Create.class) GdcDetailsDTO gdcDetailsDTO)
       throws URISyntaxException, MethodArgumentNotValidException {
     log.debug("REST request to save GdcDetails : {}", gdcDetailsDTO);
-    gdcDetailsValidator.validate(gdcDetailsDTO);
+    gdcDetailsValidator.validate(gdcDetailsDTO, null, Create.class);
     GdcDetailsDTO result = gdcDetailsService.save(gdcDetailsDTO);
     return ResponseEntity.created(new URI("/api/gdc-details/" + result.getId()))
         .headers(HeaderUtil.createEntityCreationAlert(ENTITY_NAME, result.getId().toString()))
@@ -91,7 +91,8 @@ public class GdcDetailsResource {
       @RequestBody @Validated(Update.class) GdcDetailsDTO gdcDetailsDTO)
       throws URISyntaxException, MethodArgumentNotValidException {
     log.debug("REST request to update GdcDetails : {}", gdcDetailsDTO);
-    gdcDetailsValidator.validate(gdcDetailsDTO);
+    GdcDetailsDTO originalDto = gdcDetailsService.findOne(gdcDetailsDTO.getId());
+    gdcDetailsValidator.validate(gdcDetailsDTO, originalDto, Update.class);
     if (gdcDetailsDTO.getId() == null) {
       return ResponseEntity.badRequest()
           .headers(HeaderUtil.createFailureAlert(ENTITY_NAME, "must_provide_id",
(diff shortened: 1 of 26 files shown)


JosephKelly (commented):
I know this complicates what is already a large PR but there is a risk associated with doing validation outside of the transaction.

My "extra sense" was set off by the fetching of an additional resource. After a short bit of reading, I'm convinced we should consider and document whether we should now validate in the Service layer, rather than the Resource/Controller.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/PersonValidator.java
```diff
@@ -66,15 +71,35 @@ public PersonValidator(PersonRepository personRepository, ReferenceService refer
...
    * @throws MethodArgumentNotValidException if there are validation errors
    */
-  public void validate(PersonDTO personDto) throws MethodArgumentNotValidException {
-    List<FieldError> fieldErrors = new ArrayList<>();
-    fieldErrors.addAll(checkMandatoryFields(personDto));
-    fieldErrors.addAll(checkRole(personDto));
-    fieldErrors.addAll(checkPublicHealthNumber(personDto));
+  public void validate(PersonDTO personDto, PersonDTO originalDto, Class<?> validationType)
+      throws MethodArgumentNotValidException {
+    List<FieldError> fieldErrors = new ArrayList<>(checkMandatoryFields(personDto));
```
JosephKelly, line 84: I think this is fine but want to check we've agreed/let the Data "PO" know we are still check the new `personDto` has all fields rather than the merged object having all the mandatory fields.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/GdcDetailsResource.java
```diff
@@ -91,7 +91,8 @@ public ResponseEntity<GdcDetailsDTO> updateGdcDetails(
       @RequestBody @Validated(Update.class) GdcDetailsDTO gdcDetailsDTO)
       throws URISyntaxException, MethodArgumentNotValidException {
     log.debug("REST request to update GdcDetails : {}", gdcDetailsDTO);
-    gdcDetailsValidator.validate(gdcDetailsDTO);
+    GdcDetailsDTO originalDto = gdcDetailsService.findOne(gdcDetailsDTO.getId());
+    gdcDetailsValidator.validate(gdcDetailsDTO, originalDto, Update.class);
```
JosephKelly, line 95: If refactoring, this could be placed after the id check like `ContactDetailsResource.updateContactDetails()`.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/GmcDetailsResource.java
```diff
@@ -91,7 +91,8 @@ public ResponseEntity<GmcDetailsDTO> updateGmcDetails(
       @RequestBody @Validated(Update.class) GmcDetailsDTO gmcDetailsDTO)
       throws URISyntaxException, MethodArgumentNotValidException {
     log.debug("REST request to update GmcDetails : {}", gmcDetailsDTO);
-    gmcDetailsValidator.validate(gmcDetailsDTO);
+    GmcDetailsDTO originalDto = gmcDetailsService.findOne(gmcDetailsDTO.getId());
+    gmcDetailsValidator.validate(gmcDetailsDTO, originalDto, Update.class);
```
JosephKelly, line 95: If refactoring, this could be placed after the id null check like `ContactDetailsResource.updateContactDetails()`.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/validation/ContactDetailsValidator.java
```diff
@@ -50,23 +56,46 @@ public ContactDetailsValidator(ReferenceServiceImpl referenceService,
...
+
     List<FieldError> fieldErrors = new ArrayList<>();
-    fieldErrors.addAll(checkTitle(dto, currentOnly));
-    fieldErrors.addAll(checkEmail(dto));
+    final boolean currentOnly = true;
+
+    Map<String, Function<ContactDetailsDTO, List<FieldError>>> validators = Map.of(
+        FIELD_NAME_TITLE, dto -> checkTitle(dto, currentOnly),
+        FIELD_NAME_EMAIL, this::checkEmail
+    );
```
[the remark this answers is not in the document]
JosephKelly: Same goes for the other validators.

JosephKelly (approved):
I've flagged a couple of things for a follow up PR if you agree.

PR #1075, opened 2026-05-20
feat: include programme owner in revalidation record dto

CaiWillis:
TIS21-8289


JosephKelly (changes requested):
Separate versioning within a single maven project sucks... you should update the `project.version` in the persistence and client modules (& while you're there and as a rule, update the version used by the TCS Service too).

Why? If this is merged it will silently force external services as well as TCS to use the latest version.

JosephKelly (approved):

PR #1076, opened 2026-05-22
fix: use correct field for transmitting designated body code of progr…

CaiWillis:
…amme owner to reval

TIS21-8289

### DIFF
--- a/tcs-client/pom.xml
+++ b/tcs-client/pom.xml
@@ -6,7 +6,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-client</artifactId>
-  <version>6.10.0</version>
+  <version>6.9.1</version>
   <packaging>jar</packaging>
   <dependencies>
     <dependency>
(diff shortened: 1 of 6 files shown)


tcs-client/pom.xml
```diff
@@ -6,7 +6,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-client</artifactId>
-  <version>6.9.0</version>
+  <version>6.10.0</version>
```
hihilary: I think this can be a ~~minor~~ patch version bump.

tcs-persistence/pom.xml
```diff
@@ -13,12 +13,12 @@
   <description>A module containing the model and persistence layer for TCS</description>
   <groupId>uk.nhs.tis</groupId>
   <artifactId>tcs-persistence</artifactId>
-  <version>2.37.0</version>
+  <version>2.38.0</version>
```
hihilary: Same here. ~~Minor~~ patch is enough.

tcs-service/pom.xml
hihilary: tcs-service worth a ~~minor~~ patch version bump too.

tcs-client/pom.xml
CaiWillis: Do you mean patch?

hihilary: Sorry I meant to say patch, but somehow I typed minor instead.

hihilary (approved):

PR #1078, opened 2026-06-04
chore: remove old reval rabbit config to prevent conflicts

CaiWillis:
TIS21-8285

### DIFF
--- a/tcs-service/pom.xml
+++ b/tcs-service/pom.xml
@@ -12,7 +12,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-service</artifactId>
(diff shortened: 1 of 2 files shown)


JosephKelly (commented):
It looks like there are other references to `connection.syncstart` in this repo, i.e. more that needs to be removed for consistency.

CaiWillis:
> It looks like there are other references to `connection.syncstart` in this repo, i.e. more that needs to be removed for consistency.

I can't see anything that shouldn't be there for `connection.syncstart`, we still want these queues (except `queue.connection.update`), just not to be managed by Spring Beans.

JosephKelly (approved):

PR #1080, opened 2026-06-17
feat: add necessary code for tis sync

jayanta2018:
As we agreed to split the current tcs PR and this one is for the tis sync. For PostView I was following this table:
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:w="urn:schemas-microsoft-com:office:word"
xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Word.Document>
<meta name=Generator content="Microsoft Word 15">
<meta name=Originator content="Microsoft Word 15">
<link rel=File-List
href="file:///C:/Users/NM2~1.UHB/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

<link rel=themeData
href="file:///C:/Users/NM2~1.UHB/AppData/Local/Temp/msohtmlclip1/01/clip_themedata.thmx">
<link rel=colorSchemeMapping
href="file:///C:/Users/NM2~1.UHB/AppData/Local/Temp/msohtmlclip1/01/clip_colorschememapping.xml">

<style>

</style>

</head>

<body lang=EN-GB style='tab-interval:36.0pt;word-wrap:break-word'>

Field | UI   usage | PostView   ES model | Reason
-- | -- | -- | --
nationalPostNumber | Sorting   only | String   with FieldType.Keyword | Sorting   needs exact keyword value
primarySiteId | Displayed   as Primary site text, used for filtering only | Long   primarySiteId, Decorated with primarySiteKnownAs | Filter should use ID; UI   display needs text
approvedGradeId | Displayed   as Approved grade text, used for filtering | Long   approvedGradeId, Decorated with approvedGradeCode and approvedGradeName for   display | Filter   should use ID; UI display needs text
programmeNames | Filtering | List<String>   with FieldType.Keyword | Multi-value   filter field
fundingTypes | Filtering | List<String>   with FieldType.Keyword | Multi-value   filter field
trustIds | Permission   filtering | List<Long> | Multi-value   ID filter
programmeIds | Permission   filtering | List<Long> | Multi-value   ID filter
currentTraineeSurnames   /currentTraineeForenames | Display/sorting   only | String | Display/sort field, not   permission/filter list
owner | Filtering | String   with FieldType.Keyword | Multi-value filter field
status | Filtering | Status   enum with FieldType.Keyword | Multi-value filter field
primarySpecialtyId | Filtering | FieldType.Long | Single value
primarySpecialtyCode | Filtering-   Display/searchable text | String   with FieldType.Keyword | Exact value/code
primarySpecialtyName | Filtering-Display/searchable   text | String | Exact value

</body>

</html>


tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/job/post/PostView.java
```diff
@@ -0,0 +1,73 @@
...
+  @Field(type = FieldType.Keyword)
+  private String fundingType;
+
+  @Field(type = FieldType.Keyword)
+  private String owner;
+
+  @Field(type = FieldType.Long)
+  private List<Long> trustIds;
+
+  @Field(type = FieldType.Long)
```
hihilary, line 76: Are the annotations still required for `trustIds` & `programmeIds`?

tcs-persistence/src/main/resources/queries/postViewForES.sql
```diff
@@ -0,0 +1,46 @@
...
+primarySiteId,
+GROUP_CONCAT(distinct programmeName SEPARATOR ', ') programmes,
+GROUP_CONCAT(distinct fundingType SEPARATOR ', ') fundingType,
+nationalPostNumber,
+fundingStatus,
+owner,
+GROUP_CONCAT(surnames SEPARATOR ', ') surnames,
+GROUP_CONCAT(forenames SEPARATOR ', ') forenames,
+GROUP_CONCAT(distinct trustId SEPARATOR ',') trustIds,
+GROUP_CONCAT(distinct programmeId SEPARATOR ',') programmeIds
```
hihilary: I noticed that the concatenated `trustIds` and `programmeIds` don't include a space after the separator, unlike the other concatenated values. Is there a particular reason for using a different delimiter format here?

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/job/post/PostView.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+@Document(indexName = "posts")
+@Data
+public class PostView {
+
+  @Id
+  private Long id;
+
+  private String currentTraineeSurname;
+  private String currentTraineeForenames;
```
hihilary, line 45: `currentTraineeSurnames`, `currentTraineeForenames`, `programmes`, and `fundingType` are all concatenated strings in the SQL query. Is there a reason why these are represented as concatenated strings rather than lists, like `trustIds` and `programmeIds`?

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/job/post/PostView.java
```diff
@@ -0,0 +1,73 @@
...
+  private String primarySpecialtyCode;
+  private String primarySpecialtyName;
+
+  private String programmeNames;
+
+  @Field(type = FieldType.Keyword)
+  private Status status;
+
+  @Field(type = FieldType.Keyword)
+  private String fundingType;
```
hihilary: `fundingType` is a concatenated string from multiple funding types I think. Is it okay to use Keyword type here?

tcs-persistence/src/main/resources/queries/postViewForES.sql
jayanta2018: I have replaced the SQL with the working version of the new one.

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/job/post/PostView.java
jayanta2018: currentTraineeSurnames and currentTraineeForenames are for display and sorting only. So I don't know if they are candidate for lists. 

I have changed programmes and fundingType  to lists as they will be used as Multi-value filtered fields.

trustIds  and programmeIds will be used multi-value Permission filtering

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/job/post/PostView.java
jayanta2018: Yes. Multi-value filter field and keyword is needed.

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/job/post/PostView.java
jayanta2018: They are not strictly required  but I would prefer to keep them explicit. These fields are used for permission filtering via termsQuery, so mapping them clearly as Long avoids ambiguity and makes the intended ES mapping obvious. 

I will also make the other ID fields explicit for consistency.

tcs-persistence/src/main/resources/queries/postViewForES.sql
hihilary: I don't think this comment has been addressed yet.

hihilary:
Whenever any data related to the post list is updated, we need to use the WHERE clause to identify the affected post, then rerun this query to retrieve the updated data and write it back to Elasticsearch. 

However, the current query has two limitations:
1) It cannot filter by placement.
2) Trainee names are stored as a concatenated string, making them even harder to query against.

As a result, the query cannot be used to identify posts affected by updates to placements or trainee names.

tcs-persistence/src/main/resources/queries/postViewForES.sql
jayanta2018: Good spot.
There was no particular/special reason for this. SEPARATOR ', ' -the one with space was used for display strings and SEPARATOR ','-the one with no space was used for non strings.

jayanta2018:
> Whenever any data related to the post list is updated, we need to use the WHERE clause to identify the affected post, then rerun this query to retrieve the updated data and write it back to Elasticsearch.
> 
> However, the current query has two limitations:
> 
> 1. It cannot filter by placement.
> 2. Trainee names are stored as a concatenated string, making them even harder to query against.
> 
> As a result, the query cannot be used to identify posts affected by updates to placements or trainee names.

Shall we use the old/existing query then?

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/job/post/PostView.java
```diff
@@ -0,0 +1,76 @@
...
+  private List<String> programmeNames;
+
+  @Field(type = FieldType.Keyword)
+  private Status status;
+
+  @Field(type = FieldType.Keyword)
+  private List<String> fundingTypes;
+
+  @Field(type = FieldType.Keyword)
+  private List<String> owners;
```
JosephKelly: I can't see the basis for being a `List` rather than a single post owner (Local Office).

tcs-persistence/src/main/resources/queries/postViewForES.sql
JosephKelly, line 1: I've been exploring the difference between this and the `./postView.sql` file.  If possible, it would be preferable to use the existing file or describe the reason a separate file is necessary.  This may require some query profiling to verify that additional fields don't materially impact performance on existing usage.

The person view used the same query up until it diverged too much.

JosephKelly (commented):
I started with the table in the PR description.  It's useful to have and I noticed that it doesn't include several fields; `status`, `primarySpecialty{Code|Name}`.

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/job/post/PostView.java
jayanta2018: Agreed and changed

jayanta2018:
> I started with the table in the PR description. It's useful to have and I noticed that it doesn't include several fields; `status`, `primarySpecialty{Code|Name}`.

Thanks for spotting this. I have added them.

tcs-persistence/src/main/resources/queries/postViewForES.sql
jayanta2018: Provided a justification comment.

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/job/post/PostView.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+@Document(indexName = "posts")
+@Data
+public class PostView {
+
+  @Id
+  private Long id;
+
+  private String currentTraineeSurnames;
+  private String currentTraineeForenames;
```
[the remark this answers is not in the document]
jayanta2018: Applied suggestion

tcs-persistence/src/test/java/com/transformuk/hee/tis/tcs/service/job/post/PostViewTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @Test
+  void shouldMapStatusAsKeywordField() throws NoSuchFieldException {
+    assertKeywordField("status");
+  }
+
+  @Test
+  void shouldMapFundingTypeAsKeywordField() throws NoSuchFieldException {
+    assertKeywordField("fundingTypes");
+  }
```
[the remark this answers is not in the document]
jayanta2018: Test name modified

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/job/post/PostView.java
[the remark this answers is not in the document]
JosephKelly: It doesn't look like you have.  Are you missing something from a commit?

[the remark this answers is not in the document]
jayanta2018: > It doesn't look like you have. Are you missing something from a commit?

Sorry, my bad. Applied it now. Thanks for spotting this.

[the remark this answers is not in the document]
CaiWillis: Also I'm not sure I agree with the copilot suggestion, if we want partial matches on names (e.g. a search for "John" to match "Johnathan") You'll want FieldType.Text or similar
https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/keyword
https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/text-type-family

tcs-persistence/src/main/resources/queries/postViewForES.sql
hihilary: Do we still nee this file?
I remember that we agreed to use existing postView.sql for TIS-Sync to do the Post ES Sync.

jayanta2018: Yes, Yafang, we need this file. It is lot different from the postView.sql file and the justification comments have this explanation. Without this file we won't be able to do the permission filtering.

hihilary: I think this SQL can functionally replace the original `postView.sql`. It mainly just hardcodes the trust join into the template, but there is no major functional difference.

In my opinion, there are only two differences:

- The query result will include two additional columns: "trustIds" and "programmeIds".
- For non-trust users, there will be an extra level of LEFT JOIN, which may make the query slightly slower, but for trust users the behaviour remains the same.

If I remember correctly, we‘ve already discussed this in the last huddle. cc. @JosephKelly

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/job/post/PostView.java
```diff
@@ -39,9 +39,9 @@ public class PostView {
   @Id
   private Long id;
 
-  @Field(type = FieldType.Keyword)
+  @Field(type = FieldType.Text)
   private String currentTraineeSurnames;
```
CaiWillis, line 43: Sorry Jay, I think I've confused things, these should be keyword to go with the wildcard search implemented in the other PR. I'd got it wrong, text only helps if we have a custom analyzer, which we don't

jayanta2018: Thanks changed it

tcs-service/pom.xml
```diff
@@ -80,7 +80,7 @@
     <dependency>
       <groupId>uk.nhs.tis</groupId>
       <artifactId>tcs-persistence</artifactId>
-      <version>2.37.2</version>
+      <version>2.38.0</version>
```
hihilary, line 83: The tcs-service version needs a bump.

jayanta2018: Nice spot. Thanks. 
Done

hihilary (approved):
To be honest, I'm not sure what the decision is regarding the copyright. Unless anyone has any other suggestions, I'm happy to leave it as it is.

PR #1081, opened 2026-06-22
fix: fix the post view sql

jayanta2018:
- Our existing query is slower. 

- The change is an improvement because it aggregates Placement and ContactDetails separately by postId before joining them to the main Post query. 

- Tested locally and noticed the time it takes much less.

### DIFF
--- a/tcs-persistence/src/main/resources/queries/postView.sql
+++ b/tcs-persistence/src/main/resources/queries/postView.sql
@@ -1,4 +1,4 @@
-select distinct id,
+select id,
 approvedGradeId,
 primarySpecialtyId,
 primarySpecialtyCode,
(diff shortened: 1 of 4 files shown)


jayanta2018:
Some tests failed. Checking them.

tcs-persistence/src/main/resources/queries/postView.sql
```diff
@@ -10,35 +10,43 @@ nationalPostNumber,
...
+        FROM `Placement` pl
+        LEFT JOIN `ContactDetails` c on pl.`traineeId` = c.`id` WHERE curdate() BETWEEN pl.`dateFrom` AND pl.`dateTo`
+        GROUP BY pl.`postId`
+    ) trainees on trainees.`postId` = p.`id`
+    LEFT JOIN `ProgrammePost` pp on pp.`postId` = p.`id`
     LEFT JOIN `Programme` prg on prg.`id` = pp.`programmeId`
- TRUST_JOIN
- WHERECLAUSE
+    TRUST_JOIN
+WHERECLAUSE
```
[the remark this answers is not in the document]
jayanta2018: Fixed

tcs-persistence/src/main/resources/queries/postView.sql
```diff
@@ -10,35 +10,43 @@ nationalPostNumber,
...
 ) as ot
-group by id,approvedGradeId,primarySpecialtyId,primarySpecialtyCode,primarySpecialtyName,primarySiteId,nationalPostNumber,fundingStatus,owner,intrepidId
- ORDERBYCLAUSE
- LIMITCLAUSE
-;
\ No newline at end of file
+group by id,approvedGradeId,primarySpecialtyId,primarySpecialtyCode,primarySpecialtyName,
+primarySiteId,nationalPostNumber,fundingStatus,owner,surnames,forenames,intrepidId
```
hihilary: It seems `nationalPostNumber,fundingStatus,owner,intrepidId` are not necessary in the group by clause.

tcs-persistence/src/main/resources/queries/postView.sql
```diff
@@ -1,4 +1,4 @@
-select distinct id,
+select id,
```
hihilary: Just wanted to check why `distinct` is removed

tcs-persistence/src/main/resources/queries/postView.sql
hihilary, line 50: The indention here doesn't seem to be the same as TRUST_JOIN

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
```diff
@@ -759,10 +759,10 @@ protected String createWhereClause(final String searchString,
       columnFilters.forEach(cf -> {
         switch (cf.getName()) {
           case "currentTraineeSurname":
-            whereClause.append(" AND surname in (:surnameList)");
+            whereClause.append(" AND trainees.`surnames` in (:surnameList)");
```
hihilary: To simulate this behaviour, I tried apply `WHERE 1=1 and trainee.`surnames` in ("xx")` to replace `WHERECLAUSE` in the query but got an error `Unknown column 'trainee.surnames' in 'where clause'`. I'm wondering if prefix `trainees.` is really required.

tcs-persistence/src/main/resources/queries/postView.sql
jayanta2018: I have brought back the `distinct`. Looks lik it improves the performance. Thanks for spotting this.

tcs-persistence/src/main/resources/queries/postView.sql
jayanta2018: Thanks. Indented it back.

tcs-persistence/src/main/resources/queries/postView.sql
jayanta2018: Yeah. Removed them.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
jayanta2018: The prefix is required because after the SQL refactor, surname/forenames are no longer direct columns from ContactDetails in the main query. 
They now come from the derived table aliased as `trainees`.

The alias in the SQL is `trainees`, plural, so the condition should be:

`trainees`.`surnames` in (:surnameList)
I have tested this with the following
WHERE 1=1 AND trainees.`surnames` in ('xy', 'ab') and it worked.

I think the error in the test example appears to come from using `trainee` singular rather than the actual alias `trainees`, which is plural.

tcs-persistence/src/main/resources/queries/postView.sql
```diff
@@ -10,35 +10,43 @@ nationalPostNumber,
...
-    LEFT JOIN `ContactDetails` c on pl.traineeId = c.id
-    LEFT JOIN `ProgrammePost` pp on pp.postId = p.id
+    LEFT JOIN (
+        SELECT pl.`postId`,
+        GROUP_CONCAT(DISTINCT c.`surname` SEPARATOR ', ') AS `surnames`,
+        GROUP_CONCAT(DISTINCT c.`forenames` SEPARATOR ', ') AS `forenames`
+        FROM `Placement` pl
+        LEFT JOIN `ContactDetails` c on pl.`traineeId` = c.`id` WHERE curdate() BETWEEN pl.`dateFrom` AND pl.`dateTo`
+        GROUP BY pl.`postId`
+    ) trainees on trainees.`postId` = p.`id`
```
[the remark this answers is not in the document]
jayanta2018: Good point. 
DISTINCT on surname and forenames separately can remove legitimate duplicate names where different trainees share the same surname or forename.

I updated the derived table as per your suggestions. 
Tested it locally and observed the performance improvement as well.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
hihilary: Ah, I mistakenly used "trainee". Thanks for the explaination.

hihilary: I tried both `where 1=1 and surnames in ("xx")` and `where 1=1 and trainees.surnames in ("xx")` for the WHERECLAUSE. They gave me the same results. Can you have a double check if the prefix "trainees." is still required please? Thanks.

jayanta2018: You are absoultely correct. I did check them. They work in both occassions. I have now removed the prefix.

hihilary (approved):

PR #1085, opened 2026-07-09
fix: update java version for sonar

jayanta2018:
Sonar Cloud now requires the scanner/runtime used for analysis to run on Java 21 or later. Their Maven scanner documentation lists “Java 21 or later” as a prerequisite, with Java 17 deprecated. Sonar’s general requirements also say Java 17 support for scanner runtimes ends for Sonar Cloud in July 2026.

### DIFF
--- a/.github/workflows/ci-cd-workflow.yml
+++ b/.github/workflows/ci-cd-workflow.yml
@@ -24,7 +24,7 @@ jobs:
         with:
           cache: maven
           distribution: temurin
-          java-version: 17
+          java-version: 21
 
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v5


hihilary (approved):

JosephKelly (commented):
From a quick look, I can't see why we aren't migrating to the shared workflow.

It may well be that there is bespoke config I've not picked up on but at least wanted to highlight use of the shared workflow as the preferred solution.

jayanta2018:
> From a quick look, I can't see why we aren't migrating to the shared workflow.
> 
> It may well be that there is bespoke config I've not picked up on but at least wanted to highlight use of the shared workflow as the preferred solution.

Shared repo is used for PR analysis. Yes I am changing that.

PR #1089, opened 2026-07-19
Feat/update Post ES for changes TIS21-8298

hihilary:
I've handled Post changes from below sources:
1. ContactDetails updates:  ContactDetails is saved before Person is saved. So we need to handle the updates in ContactDetails rather than Person side. 
2. ~~Person deletes~~: Person cannot be deleted if there're any other entities linked to it, eg. placements. When a placement is removed or set to INACTIVE, the person is removed from the Post View. So this is not needed.
3. ~~Person creates~~: no placements will be created with a Person create, so this is not needed.
4. Placement creates/updates/deletes
5. Post creates/updates
6. Post deletes: Post cannot be deleted if there're any other entities linked to it and there're no access to it via the UI. I've handled it, but not sure it is really needed.
7. Specialty(name) updates
8. Programme(name) updates: when a Programme is updated, it typically affects a large number of posts (potentially hundreds), so updating them all can take several minutes to complete (after the refactoring, the update is significantly faster).

TIS21-8298

----------------------------------------------------------------------------------------------------------------------------------------

For the 4 code smells identified by Sonar:
1. Remove this field injection and use constructor injection instead X 2: 
there're a lot of member parameters in those classes. I'm not really sure if I have to use constructor. Please suggest if this is needed.
~~3. Specify rollback behavior for checked exceptions using "rollbackFor" or "noRollbackFor" attributes: 
I'm not sure what this is and it's not related to my work.~~ this one is gone after merging from main
4. Make this anonymous inner class a lambda: 
this is a false positive. The annoymous inner class I added does't have an abstract method, so I cannot change it to a lambda.


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
```diff
@@ -167,6 +174,7 @@ public PostDTO save(PostDTO postDTO) {
     updateFundingStatus(post);
     PostDTO savedPostDto = postMapper.postToPostDTO(post);
     handleNewPostEsrNotification(postDTO);
+    publishPostSavedEventAfterCommit(savedPostDto);
```
hihilary: When doing the local testing. I noticed a race condition that saving a post to DB might take longer than listener event firing. So I have to make sure the event is fired after the transaction is committed.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostElasticSearchService.java
```diff
@@ -0,0 +1,129 @@
...
+
+  /**
+   * Updates the PostView document in Elasticsearch for the given postId.
+   * If the post does not exist in the database, it will be deleted from Elasticsearch.
+   *
+   * @param postId the id of the post to update
+   */
+  public synchronized void updatePostDocument(Long postId) {
+    Preconditions.checkNotNull(postId, "Person Id cannot be null");
+
```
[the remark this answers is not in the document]
hihilary: Fixed.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/listener/PostEventListener.java
```diff
@@ -0,0 +1,75 @@
...
+   * Handles the PostDeletedEvent by updating the corresponding PostView document in Elasticsearch.
+   *
+   * @param event the PostDeletedEvent containing the post ID to update
+   */
+  @EventListener
+  public void handlePostDeletedEvent(PostDeletedEvent event) {
+    Long postId = event.getPostId();
+    LOG.info("Received PlacementDeleteEvent for placement id [{}]", postId);
+    postElasticSearchService.updatePostDocument(postId);
+  }
```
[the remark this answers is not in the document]
hihilary: Fixed.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/listener/person/ContactDetailsElasticSearchEventListener.java
```diff
@@ -0,0 +1,91 @@
...
+    LOG.info("Received ContactDetails saved event for id [{}]",
+        event.getContactDetailsDto().getId());
+
+    // When there's name change, get the person's current placements,
+    // and send the postId to PostElasticSearchService for post updates
+    ContactDetailsDTO oldContactDetailsDto = event.getPreviousContactDetailsDto();
+    if (oldContactDetailsDto != null
+        && (!oldContactDetailsDto.getForenames().equals(newContactDetailsDto.getForenames())
+            || !oldContactDetailsDto.getSurname().equals(newContactDetailsDto.getSurname()))) {
+
```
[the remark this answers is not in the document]
hihilary: Fixed.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ProgrammeServiceImpl.java
```diff
@@ -97,13 +104,24 @@ public ProgrammeDTO update(ProgrammeDTO programmeDTO) {
...
   public List<ProgrammeDTO> save(List<ProgrammeDTO> programmeDtos) {
     log.debug("Request to save {} programmes.", programmeDtos.size());
-    List<Programme> programmes = programmeMapper.programmeDTOsToProgrammes(programmeDtos);
 
+    Map<Long, ProgrammeDTO> existingProgrammeDtos = programmeRepository.findByIdIn(
+            programmeDtos.stream().map(ProgrammeDTO::getId).collect(Collectors.toSet()))
+        .stream()
+        .collect(Collectors.toMap(Programme::getId, programmeMapper::programmeToProgrammeDTO,
+            (existing, replacement) -> existing, HashMap::new));
+
```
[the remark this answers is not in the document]
hihilary: Fixed.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ContactDetailsServiceImpl.java
```diff
@@ -43,23 +59,48 @@ public ContactDetailsServiceImpl(ContactDetailsRepository contactDetailsReposito
...
-    log.debug("Request to save ContactDetails : {}", contactDetailsDTOs);
-    List<ContactDetails> contactDetails = contactDetailsMapper.toEntity(contactDetailsDTOs);
+  public List<ContactDetailsDTO> save(List<ContactDetailsDTO> contactDetailsDtos) {
+    log.debug("Request to save ContactDetails : {}", contactDetailsDtos);
+
+    Map<Long, ContactDetailsDTO> existingContactDetailDtos = contactDetailsRepository.findAllById(
+            contactDetailsDtos.stream().map(ContactDetailsDTO::getId).collect(Collectors.toSet()))
+        .stream()
+        .collect(Collectors.toMap(ContactDetails::getId, contactDetailsMapper::toDto,
+            (existing, replacement) -> existing, HashMap::new));
```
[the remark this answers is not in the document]
hihilary: Fixed.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ContactDetailsServiceImpl.java
```diff
@@ -43,23 +59,48 @@ public ContactDetailsServiceImpl(ContactDetailsRepository contactDetailsReposito
...
+        existingContactDetails);
+
     ContactDetails contactDetails = contactDetailsMapper.toEntity(contactDetailsDTO);
     contactDetails = contactDetailsRepository.saveAndFlush(contactDetails);
-    return contactDetailsMapper.toDto(contactDetails);
+    ContactDetailsDTO updatedContactDetailsDto = contactDetailsMapper.toDto(contactDetails);
+
+    applicationEventPublisher.publishEvent(
+        new ContactDetailsSavedEvent(existingContactDetailsDto, updatedContactDetailsDto));
+    return updatedContactDetailsDto;
```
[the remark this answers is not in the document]
hihilary: Fixed.

tcs-persistence/src/main/java/com/transformuk/hee/tis/tcs/service/repository/PlacementRepository.java
```diff
@@ -100,6 +100,15 @@ List<Placement> findCurrentPlacementForTrainee(
...
       @Param("placementTypes") List<String> placementTypes);
 
+  List<Placement> findByTraineeIdAndDateFromLessThanEqualAndDateToGreaterThanEqual(
+      Long traineeId, LocalDate currentDateFrom, LocalDate currentDateTo);
+
+  default List<Placement> findAllCurrentPlacementsForTrainee(
+      Long traineeId, LocalDate currentDate) {
+    return findByTraineeIdAndDateFromLessThanEqualAndDateToGreaterThanEqual(
+        traineeId, currentDate, currentDate);
+  }
```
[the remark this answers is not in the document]
hihilary: Fixed.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/listener/person/PlacementElasticsearchEventListenerTest.java
```diff
@@ -33,30 +38,119 @@ public class PlacementElasticsearchEventListenerTest {
   @Mock
   RevalidationService revalidationService;
 
+  @Mock
+  PostElasticSearchService postElasticSearchService;
+
+  @Mock
+  Clock clock;
```
JosephKelly: Rather than mocking the clock. There's less processing involved in using an object:

```java
private static final Clock CLOCK = Clock.fixed(CURRENT_DATE.atStartOfDay(UTC).toInstant(), UTC);
```

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/listener/person/PlacementElasticSearchEventListener.java
```diff
@@ -47,10 +82,27 @@ public void handlePlacementSavedEvent(PlacementSavedEvent event) {
...
+  private boolean isCurrentPlacement(PlacementDTO placementDto) {
+    LocalDate dateFrom = placementDto.getDateFrom();
+    LocalDate dateTo = placementDto.getDateTo();
+    if (dateFrom == null || dateTo == null) {
+      return false;
+    }
+
+    LocalDate currentDate = LocalDate.now(clock);
+    return !currentDate.isBefore(dateFrom) && !currentDate.isAfter(dateTo);
   }
```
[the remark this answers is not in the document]
JosephKelly: @copilot Where is `Placement.dateTo` defined as nullable?

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/listener/person/PlacementElasticSearchEventListener.java
```diff
@@ -33,11 +58,21 @@ public class PlacementElasticSearchEventListener {
...
+        revalidationService.buildTcsConnectionInfo(placementDto.getTraineeId())
     );
+
+    PlacementDTO previousPlacementDto = event.getPreviousPlacementDto();
+    if (isCurrentPlacement(placementDto)) {
+      Long postId = placementDto.getPostId();
+      postElasticSearchService.updatePostDocument(postId);
+    } else if (previousPlacementDto != null && isCurrentPlacement(previousPlacementDto)) {
+      Long postId = previousPlacementDto.getPostId();
+      postElasticSearchService.updatePostDocument(postId);
```
JosephKelly: Quite a minor thing but instantiating a new variable is inconsistent with the use of `placementDto.getTraineeId()` above.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/listener/ProgrammeElasticSearchEventListener.java
JosephKelly, line 1: This seems like a good candidate for something else that sits firmly outside of the `...listener.person` package.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/listener/SpecialtyElasticSearchEventListener.java
JosephKelly, line 1: This seems like a good candidate for something else that sits firmly outside of the `...listener.person` package.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
JosephKelly: Have you changing the affected listeners to be a `@TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)`?
This is a potential refactor in a following PR rather than a necessary change.  The helper class looks like a good solution although I'm not familiar enough with the depth of transaction lifecycle you use.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ContactDetailsServiceImpl.java
```diff
@@ -43,23 +61,54 @@ public ContactDetailsServiceImpl(ContactDetailsRepository contactDetailsReposito
...
+        .map(ContactDetailsDTO::getId)
+        .filter(java.util.Objects::nonNull)
+        .collect(Collectors.toSet());
+
+    Map<Long, ContactDetailsDTO> existingContactDetailDtos = contactDetailIds.isEmpty()
+        ? java.util.Collections.emptyMap()
+        : contactDetailsRepository.findAllById(contactDetailIds)
+            .stream()
+            .collect(Collectors.toMap(ContactDetails::getId, contactDetailsMapper::toDto,
+                (existing, replacement) -> existing, HashMap::new));
```
JosephKelly: The use of `HashMap::new` as a factory function seems very specific.  It doesn't necessarily need to change but less code usually helps maintainability.
The merge function seems more specific than we need to be in this context.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PlacementServiceImpl.java
```diff
@@ -457,10 +463,13 @@ public PlacementDetailsDTO saveDetails(final PlacementDetailsDTO placementDetail
     placement.setSpecialties(new HashSet<>());
 
     Placement savedPlacement = placementRepository.saveAndFlush(placement);
-    PlacementDTO placementDTO = convertPlacementWithSupervisors(savedPlacement);
-    applicationEventPublisher.publishEvent(new PlacementSavedEvent(placementDTO));
+    convertPlacementWithSupervisors(savedPlacement);
 
-    return createDetails(placementDetailsDTO, placement);
+    PlacementDetailsDTO updatedPlacementDetailsDto =
+        self.createDetails(placementDetailsDTO, placement);
```
JosephKelly, line 469: Not clear at the moment why this is necessary.  I see it is related to a code smell but unclear what and why we've opted to address it by stretching dependency injection.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
```diff
@@ -431,22 +442,24 @@ public PostDTO update(PostDTO postDTO) {
...
    */
-  protected void updateFundingStatus(Post post) {
+  protected Post updateFundingStatus(Post post) {
     if (post != null) {
       Status fundingStatus = getFundingStatusForPost(post);
       post.setFundingStatus(fundingStatus);
-      postRepository.save(post);
+      return postRepository.save(post);
     }
+    return null;
```
JosephKelly: While this theoretically happens implicitly elsewhere, it seems odd to return `null` rather than throw an Exception.
This unexpected flow seems to add / cascade failure by allowing the PostFundingEventListener to publish events with a PostDTO that is `null`.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/mapper/PostViewRowMapper.java
```diff
@@ -0,0 +1,112 @@
...
+  private String getNullableString(ResultSet rs, String columnName) throws SQLException {
+    return rs.getString(columnName);
+  }
+
+  private List<Long> toLongList(String value) {
+    if (StringUtils.isEmpty(value)) {
+      return Collections.emptyList();
+    }
+
+    return Arrays.stream(value.split(","))
```
JosephKelly, line 95: Using separators consistently would be preferable.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostService.java
```diff
@@ -97,7 +97,7 @@ public interface PostService {
    *
    * @param postId        the id of the post to update
```
JosephKelly, line 98: I'm surprised copilot or sonar didn't pick up on the change to the Javadoc that should have been made here!
It's worth documenting here that it can be null if you are sticking with that implementation.

tcs-api/pom.xml
```diff
@@ -6,7 +6,7 @@
 
   <groupId>com.transformuk.hee.tis</groupId>
   <artifactId>tcs-api</artifactId>
-  <version>6.30.1</version>
+  <version>6.30.2</version>
```
JosephKelly: If remaining the same after addressing the PR comment, the change within this PR would be a good example of a minor version change in the API rather than a patch change.

tcs-client/pom.xml
```diff
@@ -37,7 +37,7 @@
     <dependency>
       <groupId>com.transformuk.hee.tis</groupId>
       <artifactId>tcs-api</artifactId>
-      <version>6.30.1</version>
+      <version>6.30.2</version>
```
JosephKelly: As good practice, changing this dependency will expose additional functionality to clients and in the future we may expect all dependency changes to be accompanied by a patch version change.

JosephKelly (dismissed):
I'm stopping short of reviewing the tests for now.  It will likely need to be picked up by someone else with any changes from this peer review.

Of the new and modified Event Listeners, there's a mix of sending post ids to update and then updating for a connected id `specialtyId`.

This PR changes the API in a backward-compatible but unusual way.  `ProgrammeDTO`s are returned with `postIds` rather than `posts`.  I would like a better understanding of why programme and specialy name changes are handled in different ways.

Most comments are suggestions for a refactor after this is merged.  As the `programmeDTO.postIds` is an API change, it would be better to address that with comments or changes before merging.
Ideally, API changes mean at least a minor version change.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/listener/person/PlacementElasticSearchEventListener.java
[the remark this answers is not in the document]
JosephKelly: @copilot Where is Placement.dateTo defined as nullable?

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PlacementServiceImpl.java
hihilary: Yes, the change purely came from a Sonar code smell, which is: java:S6809. And details are as below:
```
Methods with Spring proxy should not be called via "this"
A method annotated with Spring’s @Async, @Cacheable or @Transactional annotations will not work as expected if invoked directly from within its class.
This is because Spring generates a proxy class with wrapper code to manage the method’s asynchronicity (@Async), to cache methods invocations (@Cacheable), or to handle the transaction (@Transactional). However, when called using this, the proxy instance is bypassed, and the method is invoked directly without the required wrapper code.
```

hihilary: Should I keep the current change or roll it back?

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/mapper/PostViewRowMapper.java
hihilary: This file is a copy from the PostViewRowMapper from Sync service, because both service need to get data from the DB in the same way and insert into ES.
I agree that we should use the same separator. However, that would also require changes to both postView.sql and the Sync Service. @jayanta2018 how do you think?

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/listener/person/PlacementElasticSearchEventListener.java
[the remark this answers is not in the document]
hihilary: I've just had a check on the data and UI. Currently we don't allow null value in both dateFrom and dateTo on Admins-UI, and DB only contains some legacy data with null dateTo. So I'm going to roll back my change on this Copilot comment.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
hihilary: I don't think there's any real difference between the two approaches from a transaction perspective. I started from the Post save method, and at the time I didn't want to affect any other callers, so I made the change on the publisher side.
Actually, @Transactional is applied at the class level, so changing the listener is perfectly reasonable as well. I've now switched to using `@TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)`.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/listener/person/PlacementElasticsearchEventListenerTest.java
hihilary: Fixed. Thanks.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/listener/person/PlacementElasticSearchEventListener.java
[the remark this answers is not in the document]
hihilary: I've reverted the changes. Thanks.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/listener/person/PlacementElasticSearchEventListener.java
hihilary: I'm not sure I fully understand this comment. I've updated it based on my understanding, but please let me know if I missed anything. Thanks!

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/listener/ProgrammeElasticSearchEventListener.java
hihilary: Moved outside of `...listener.person` package. Thanks.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/listener/SpecialtyElasticSearchEventListener.java
hihilary: Moved outside of `...listener.person` package. Thanks.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ContactDetailsServiceImpl.java
hihilary: Fixed. Thanks.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/PostServiceImpl.java
hihilary: I added exception throwing on the caller of this method, and made the parameter to be @Nonnull.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostService.java
hihilary: Added the comment for the return value. Thanks.

tcs-api/pom.xml
hihilary: ~~Bumped the minor version. Thanks.~~ As I have removed `postId` from ProgrammeDTO, this version bump is no longer needed.

tcs-client/pom.xml
hihilary: Reverted the version bump of tcs-client. Thanks.

hihilary:
> I'm stopping short of reviewing the tests for now. It will likely need to be picked up by someone else with any changes from this peer review.
> 
> Of the new and modified Event Listeners, there's a mix of sending post ids to update and then updating for a connected id `specialtyId`.
> 
> This PR changes the API in a backward-compatible but unusual way. `ProgrammeDTO`s are returned with `postIds` rather than `posts`. I would like a better understanding of why programme and specialy name changes are handled in different ways.
> 
> Most comments are suggestions for a refactor after this is merged. As the `programmeDTO.postIds` is an API change, it would be better to address that with comments or changes before merging. Ideally, API changes mean at least a minor version change.

The reason I chose to add a new field to ProgrammeDTO is that I noticed the Programme entity already has the posts available. I thought we could avoid an additional database query using the long postView.sql query, which would save some processing time.

The reason I mapped only the IDs instead of returning the full DTOs is that I found some programmes have a large number of linked posts. Returning all of those details directly could potentially cause performance issues on the UI side, or the list of DTOs could be truncated due to the payload size. Therefore, I only mapped the post IDs to the DTO.

If you see this differently, happy to discuss and refactor. Thanks!

--------------------Update------------------------
I realised that using `postIds` in `ProgrammeDTO` doesn't actually save any processing time. In fact, because it requires calling `updatePostDocument` multiple times, it ends up executing that long query multiple times.
Querying by programme id, on the other hand, allows us to retrieve all affected PostViews in a single query. Based on that, I've removed `postIds` from ProgrammeDTO.
The code has been updated accordingly.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/mapper/PostViewRowMapper.java
jayanta2018: This separator is used for lists of longs for example, programmeIds, trustIds etc. These are not exposed to the FE.

jayanta2018 (approved):
Since local tests works succesfully, I am happy to approve this pull request.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/impl/ContactDetailsServiceImpl.java
hihilary: Jay and I had a double check on this `@TransactionalEventListener`, and noticed it was not fired at all when there were Post updates/creates or ContactDetails updates.
I don't know why the first time I tested with it successfully. It was might be because I didn't restart the service, etc., but really have no idea. 
So I decided to go back to the previous implementation which is pushing after commit. Thanks.

PR #1090, opened 2026-07-29
feat: validate funding subtype for funding types on save post

CaiWillis:
TIS21-8842: Back end TCS


CaiWillis:
Includes check for relevant funding subtypes as well as missing funding subtypes.

Based on patch containing refactors to integration tests

PR #1091, opened 2026-07-30
fix: add case insensitive npn search

jayanta2018:
User raised an issue on searching NPN with lower case value. This tweak will help them to achieve this. 

Some points for our future improvement:
- We had to use lower version of client elasticsearch and our version is maintained by Spring boot version 2.3.4
- caseInsensitive(true) option is available in higher version of ES. So we have to upgrade Spring boot soon and hence the Java version.

NO-CARD

### DIFF
--- a/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostElasticSearchService.java
+++ b/tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostElasticSearchService.java
@@ -234,8 +234,9 @@ public class PostElasticSearchService {
       searchQuery = StringUtils.remove(searchQuery, '"');
       String wildcard = "*" + searchQuery + "*";
 
+      addNationalPostNumberWildcardQuery(shouldQuery, searchQuery);
+
       shouldQuery
-          .should(new WildcardQueryBuilder(NATIONAL_POST_NUMBER, wildcard))
           .should(new WildcardQueryBuilder(PROGRAMME_NAMES, wildcard))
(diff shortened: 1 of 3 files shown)


tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/service/PostElasticSearchService.java
```diff
@@ -234,8 +234,9 @@ private BoolQueryBuilder applyTextBasedSearchQuery(String searchQuery) {
       searchQuery = StringUtils.remove(searchQuery, '"');
       String wildcard = "*" + searchQuery + "*";
 
+      addNationalPostNumberWildcardQuery(shouldQuery, searchQuery);
+
       shouldQuery
-          .should(new WildcardQueryBuilder(NATIONAL_POST_NUMBER, wildcard))
           .should(new WildcardQueryBuilder(PROGRAMME_NAMES, wildcard))
           .should(new MatchQueryBuilder(PRIMARY_SPECIALTY_NAME, searchQuery))
           .should(new WildcardQueryBuilder(FUNDING_TYPES, wildcard))
```
CaiWillis, line 242: Just for clarity, were you planning on adding the additional name match queries: 

```
.should(new MatchQueryBuilder(CURRENT_TRAINEE_FORENAMES, searchQuery))
.should(new MatchQueryBuilder(CURRENT_TRAINEE_SURNAMES, searchQuery))
```

in a separate PR?

CaiWillis: Discussion in standup was to fix this separately

CaiWillis (approved):

PR #1093, opened 2026-07-30
refactor(test): merge similar tests

JosephKelly:
Using ParameterizedTests reduces the LoC to work through.

This is one of several options.  Might also be reasonable to pass through an offset & re-use the existing `TODAY` constant.

TIS21-8842, in
TIS21-8840: Require subtype for applicable funding types


JosephKelly:
The target branch has been approved for merging into main.

PR #1094, opened 2026-08-04
feat: validate person document uploads in a way that is consistent wi…

CaiWillis:
…th UI

TIS21-8872

- Added a new utility class for validating files
- File validation is ~a 3 step process: validating the file extension, checking the file "magic byte" signiture (as file extension and mime type can be spoofed) and finally~ using the TIKA detect() method, which includes a final check for mime type
- The TIKA detect() method _does_ include magic byte detection, ~however it requires an additional package to guarantee that a check for magic bytes is on the verification chain, and .xlsx and .docx share the signature of .zip files, so the explicit magic bytes check has been included (also acts as a faster fail-fast verification step) - we could bin off the use of TIKA to improve performance~
- Had to format the DocumentResourceIntTest so the diff is large, actual changes are on lines 152-173


tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/DocumentResourceTest.java
```diff
@@ -35,9 +35,13 @@
...
 
   private static final long PERSON_BASE_ID = 10;
-  private static final byte[] TEST_FILE_CONTENT = "DataDataDataDataDataData".getBytes();
-  private static final String TEST_FILE_NAME = "document.txt";
-  private static final String TEST_FILE_CONTENT_TYPE = "text/plain";
+  // Valid PDF header bytes
+  private static final byte[] TEST_FILE_CONTENT = new byte[]{
+      (byte) 0x25, (byte) 0x50, (byte) 0x44, (byte) 0x46,
+      (byte) 0x2D, (byte) 0x31, (byte) 0x2E, (byte) 0x34
+  };
```
JosephKelly: IMO, `"%PDF-1.4".getBytes();` is simpler and may render the comment unnecessary.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/FileValidator.java
```diff
@@ -0,0 +1,166 @@
...
+      "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"                  // .xlsx
+  );
+
+  // Magic bytes (file signatures) for valid file types
+  private static final byte[] PDF_SIGNATURE = {(byte) 0x25, (byte) 0x50, (byte) 0x44,
+      (byte) 0x46}; // PDF
+  private static final byte[] OLE2_SIGNATURE = {(byte) 0xD0, (byte) 0xCF, (byte) 0x11,
+      (byte) 0xE0}; // OLE2 (.doc, .xls)
+  private static final byte[] ZIP_SIGNATURE = {(byte) 0x50, (byte) 0x4B, (byte) 0x03,
+      (byte) 0x04}; // PK  (.docx, .xlsx)
```
JosephKelly: Seems like a good candidate for an enumeration.
See https://jakarta.ee/specifications/platform/11/apidocs/jakarta/activation/mimetypesfiletypemap for an existing framework that we might be able to plug into.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/FileValidator.java
```diff
@@ -0,0 +1,166 @@
...
+        return false;
+      }
+    } catch (IOException ex) {
+      LOG.error("Failed to validate file signature", ex);
+      return false;
+    }
+
+    // MIME type check via Tika
+    try {
+      final String detectedType = TIKA.detect(documentParam.getInputStream(), filename);
```
JosephKelly: If this provides useful benefits, e.g. validation beyond what is available from Spring, then adding the additional dependency makes sense.

I didn't quite get to checking the docs for whether Spring `MetaData` relies just on the Content-Type header

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/FileValidator.java
```diff
@@ -0,0 +1,166 @@
...
+      return "";
+    }
+
+    final int lastDotIndex = filename.lastIndexOf('.');
+    if (lastDotIndex <= 0 || lastDotIndex == filename.length() - 1) {
+      return "";
+    }
+
+    return filename.substring(lastDotIndex + 1);
+  }
```
JosephKelly: I'd suggest `org.springframework.util.StringUtils#getFilenameExtension`.

N.B. will require some refactoring, e.g. checking `StringUtils#isEmpty` on the return value.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/FileValidationUtil.java
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  /**
+   * Validates the file type: 1. Extension check 2. Magic bytes check (file signature) 3. MIME type
+   * check
+   *
+   * @param documentParam the file to validate
+   * @return true if all validation layers pass, else false
+   */
+  public static boolean isValidFileType(final MultipartFile documentParam) {
```
JosephKelly, line 29: I'd personally prefer a `Validator` to follow naming & behaviour we get with classes like those in the package `com.transformuk.hee.tis.tcs.service.api.validation`, i.e. deciding whether it's better to throw a `MethodArgumentNotValidException`, TCS `ValidationException` or `jakarta.validation.ValidationException`.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/DocumentResource.java
```diff
@@ -377,6 +382,18 @@ private Optional<DocumentDTO> createDocument(final MultipartFile documentParam,
     return Optional.of(document);
   }
 
+  private boolean isValidFileType(final MultipartFile documentParam) {
+    final String filename = documentParam.getOriginalFilename();
```
JosephKelly: The filename is extracted in the method annotated with `@PostMapping`.  I'd consider refactoring this chain of 2 nested method calls with single usage.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/DocumentResourceIntTest.java
```diff
@@ -69,9 +69,21 @@ public class DocumentResourceIntTest {
...
   private static final long DOCUMENT_BASE_ID = 100;
   private static final long TAG_BASE_ID = 1000;
-  private static final byte[] TEST_FILE_CONTENT = "DataDataDataDataDataData".getBytes();
-  private static final String TEST_FILE_NAME = "document.txt";
-  private static final String TEST_FILE_CONTENT_TYPE = "text/plain";
+  // Content begins with PDF magic bytes
+  private static final byte[] TEST_FILE_CONTENT = new byte[]{
+      (byte) 0x25, (byte) 0x50, (byte) 0x44, (byte) 0x46,
+      (byte) 0x2D, (byte) 0x31, (byte) 0x2E, (byte) 0x34
+  };
```
JosephKelly, line 76: As above, I prefer the string version for readability but will go with your choice.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/util/FileValidationUtilTest.java
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  @Test
+  void isValidFileType_shouldReturnTrue_whenValidPdf() {
+    final MockMultipartFile file = new MockMultipartFile(
+        "file", "document.pdf", "application/pdf", PDF_BYTES);
+    assertThat(FileValidator.isValidFileType(file)).isTrue();
+  }
+
+  @ParameterizedTest(name = "[{index}] {0}")
```
JosephKelly, line 34: Nice!

JosephKelly (changes requested):

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/DocumentResourceTest.java
CaiWillis: (Discussed elsewhere) unfortunately other headers use strange characters making the string versions no simpler so sticking with bytes

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/DocumentResource.java
CaiWillis: I didn't like the idea of modifying the interface just for this, do you have a more specific idea you can point me to?

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/DocumentResourceTest.java
JosephKelly: I think the reply may be intended for a different comment: https://github.com/Health-Education-England/TIS-TCS/pull/1094#discussion_r3712188826.

It has been replaced with `FileSignature.PDF.bytes()`, which as use of an enum value would be, is much clearer.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/DocumentUploadConstraints.java
```diff
@@ -0,0 +1,38 @@
...
+      "doc", FileSignature.OLE2,
+      "xls", FileSignature.OLE2,
+      "docx", FileSignature.ZIP,
+      "xlsx", FileSignature.ZIP
+  );
+
+  private DocumentUploadConstraints() {
+    throw new UnsupportedOperationException("Utility class cannot be instantiated");
+  }
+}
```
JosephKelly: It would be good to understand your preference of this class over an enum.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/FileValidator.java
JosephKelly: I probably could have been clearer that I was selecting the whole block as something that could form an enum, i.e.
with a constructor `AllowedFileType(String extenstion, String mimeType, List<Byte> signature)`.

If we are able to make use of the mimetypesfilemap or Tika's mimetypes and file signatures, we'd have to give further consideration to if & what shape of enum to use.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/FileValidator.java
JosephKelly: Please confirm the validation outside org.springframework.util.StringUtils#getFilenameExtension is necessary.  I hoped we could further simplify, if not completely replace this method.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/DocumentResource.java
JosephKelly: As far as I can see you've addressed this comment.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/FileSignature.java
JosephKelly, line 1: I think there is value in using an authoritative source for this information.
My expectation is it prevents conversations such as:

JosephKelly: I notice that from the original PR, the pdf signature has changed but it's not clear if that was as a result of the conversation about what we need to validate.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/DocumentResource.java
```diff
@@ -236,6 +241,14 @@ public ResponseEntity<DocumentId> uploadDocument(
...
         personId, filename);
 
+    try {
+      documentUploadValidator.validate(documentParam);
+    } catch (ValidationException ex) {
+      LOG.warn("Document with person '{}' and document name '{}' failed file type validation",
+          personId, filename);
+      return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
+    }
+
```
JosephKelly: It's simpler to modify the signature so it `... throws IOException, ValidationException ...` and remove the try/catch.
The response body returned to the client will contain the validation error(s).

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/util/FileValidationUtilTest.java
JosephKelly, line 1: Most of the tests look very similar.  I would probably suggest validating that we log the correct decisions with `CapturedOutput`.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/DocumentResourceTest.java
```diff
@@ -53,8 +57,11 @@ public class DocumentResourceTest {
 
   @Before
   public void setup() {
-    DocumentResource testObj = new DocumentResource(documentServiceMock, tagServiceMock);
-    mockMvc = MockMvcBuilders.standaloneSetup(testObj).build();
+    DocumentResource testObj = new DocumentResource(documentServiceMock, tagServiceMock,
+        new DocumentUploadValidator());
+    mockMvc = MockMvcBuilders.standaloneSetup(testObj)
+        .setControllerAdvice(new ExceptionTranslator())
```
JosephKelly: I can't see why you'd need this here.  It might be needed if changing the upload endpoint to throw the ValidationException.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/DocumentUploadFileType.java
```diff
@@ -0,0 +1,92 @@
...
+   *
+   * @param extension the file extension (without the dot)
+   * @return the DocumentUploadFileType for the given extension, or empty if not found
+   */
+  public static Optional<DocumentUploadFileType> fromExtension(final String extension) {
+    if (extension == null) {
+      return Optional.empty();
+    }
+    return Optional.ofNullable(BY_EXTENSION.get(extension.toLowerCase(Locale.ROOT)));
+  }
```
JosephKelly: This is close to `Enum.valueOf()`, with the exception of using an `IllegalArgumentException`. The case-insensitive match on the file extension could use the enum name.

```java
return Arrays.stream(values())
    .filter(e -> e.name().equalsIgnoreCase(extension))
    .findAny().orElseThrow(IllegalArgumentException::new)
```


It's odd that this produces an `Optional.empty()` when the only production usage of this method defaults the value back to null.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/DocumentUploadFileType.java
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Returns all allowed upload MIME types.
+   *
+   * @return the set of allowed MIME types
+   */
+  public static Set<String> allowedMediaTypes() {
+    return Arrays.stream(values())
+        .map(DocumentUploadFileType::mediaType)
+        .collect(Collectors.toUnmodifiableSet());
+  }
```
JosephKelly, line 88: Not explicitly requesting a change here but these methods instantiate new sets every time a document is uploaded.  This enum could create a static instance of each of them like ` Map<String, DocumentUploadFileType> BY_EXTENSION` to be returned by the getters.

JosephKelly (commented):
I've dedicated far too much headspace to the use of locale for it to be here without a clear idea of why this PR breaks from the existing assumption that the default locale is sufficient.

I've left some earlier comments unresolved where I think they need further comment or changes.

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/DocumentResource.java
CaiWillis: This has been done 👍

JosephKelly: My mistake.  I missed this and failed to make the comment on what was added to the `DocumentResourceIntTest`.  I thought adding the `ExceptionTranslator()` was enough, without https://github.com/Health-Education-England/TIS-TCS/pull/1094/changes#diff-fd8f00ff5fc5253bedc2303a1e7ac2aaa9d175f1a32433633978cf536a653b79R1186-R1195

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/FileSignature.java
CaiWillis: Byte file signiatures have been removed except for unit tests. PR now uses Tika.detect only, no convenient authoritative source was found

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/FileValidator.java
CaiWillis: StringUtils#getFilenameExtension is now used in place of the above

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/DocumentUploadFileType.java
CaiWillis: Makes sense and was an easy little performance/memory boost, so done

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/util/FileValidationUtilTest.java
CaiWillis: Parameterized and removed some redundant test cases

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/DocumentUploadFileType.java
CaiWillis: Done, much tidier

tcs-service/src/main/java/com/transformuk/hee/tis/tcs/service/api/util/FileValidator.java
JosephKelly: The last commit I see leaves `FileValidationUtil#extractFileExtension` in place.
I've tried thinking about why it's better to return an empty string.
I can't see a differential handling of `null` vs. `""`.
If I've missed it, this method performs 2 checks that Spring's `StringUtils#getFilenameExtension` performs.  The only current & future uses could be replaced with, e.g.

```java
final String fileExtension = Optional.ofNullable(StringUtils.getFilenameExtension(filename)).orElse("");
```

JosephKelly: Granted we'd do a bit more than necessary in the exceptional case that the filename is null if we skipped even that and went straight to

```java
expectedType = DocumentUploadFileType.fromExtension(StringUtils.getFilenameExtension(filename))
```

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/DocumentResourceIntTest.java
JosephKelly: OK, we'll leave these as is.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/DocumentResourceIntTest.java
```diff
@@ -1143,4 +1183,14 @@ public Iterator<T> iterator() {
...
+
+  @ControllerAdvice
+  private static class ValidationExceptionHandler {
+
+    @ExceptionHandler(ValidationException.class)
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    public void handleValidationException() {
+      // Status-only handler for upload validation failures in standalone MockMvc setup.
+    }
+  }
```
JosephKelly: Sorry it took a while to get round to testing real behaviour.
There are some fundamental discrepancies  in this test class compared to real behaviour, returning HTTP 400 when the deployed code returns HTTP 500 which goes undetected because it doesn't appropriately use an ExceptionHandler.

I think it would be better to identify those, even if accepting them by marking the tests to `@Ignore` them for this PR.

JosephKelly (changes requested):
Thanks for addressing the earlier comment.
I'm requesting that we limit the ability for prod code to get further broken, i.e. currently changing the `ExceptionTranslator` to return a `@ResponseStatus(HttpStatus.FORBIDDEN)` doesn't break the test you are adding.

tcs-service/src/test/java/com/transformuk/hee/tis/tcs/service/api/DocumentResourceIntTest.java
```diff
@@ -108,32 +119,60 @@ public class DocumentResourceIntTest {
...
 
     mockMvc.perform(fileUpload(DocumentResource.PATH_API + DocumentResource.PATH_DOCUMENTS)
-        .file(mockFile)
-        .contentType(MediaType.MULTIPART_FORM_DATA))
+            .file(mockFile)
+            .contentType(MediaType.MULTIPART_FORM_DATA))
         .andExpect(status().isBadRequest());
   }
 
+  @Ignore("TODO: resolve conflict with expected result and controller advice")
```
CaiWillis, line 146: Not sure if putting TODO's in an `@Ignore` is cheeky

